PR #1, opened 2020-03-02
FELIX-6232 - adding option to install parallel version of bundle

DominikSuess:


bosschaert:
Hi @DominikSuess the change looks good to me.

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
```diff
@@ -1642,8 +1648,10 @@ private void installBundle( String location, File bundleFile, int startLevel, bo
                 Bundle[] bundles = BundleContextUtil.getWorkingBundleContext(this.getBundleContext()).getBundles();
                 for ( int i = 0; i < bundles.length; i++ )
                 {
+                    boolean isSameBSN = (bundles[i].getSymbolicName() != null && bundles[i].getSymbolicName().equals( symbolicName ));
```
kwin: I think you can just do `symbolicName.equals(bundles[i].getSymbolicName())` as `symbolicName` can't be `null` here!

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
```diff
@@ -1710,6 +1718,43 @@ private String getSymbolicName( File bundleFile )
         // fall back to "not found"
         return null;
     }
+    
+    private String getBundleVersion( File bundleFile )
+    {
+        JarFile jar = null;
+        try
+        {
+            jar = new JarFile( bundleFile );
```
kwin: try with resources (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)?

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
amitjoy: BSN cannot be null according to Felix implementation but according to Javadoc, it can. I agree completely that the null check is superfluous here.

kwin: It can't be null due to the if condition at https://github.com/apache/felix-dev/pull/1/files#diff-b66e715ba83980920efaa568f6819c45L1628!

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
```diff
@@ -1617,14 +1622,15 @@ private FileItem getParameter( Map params, String name )
...
 
-    private void installBundle( String location, File bundleFile, int startLevel, boolean start, boolean refreshPackages )
+    private void installBundle( String location, File bundleFile, int startLevel, boolean start, boolean refreshPackages, boolean parallelVersion)
             throws IOException
     {
         if ( bundleFile != null )
         {
 
             // try to get the bundle name, fail if none
             String symbolicName = getSymbolicName( bundleFile );
```
cziegeler: I suggest to get bsn and version in one method, otherwise you open the file twice

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
jsedding: You could use `Objects.equals`, it handles `null` arguments.

DominikSuess: I appreciate the suggestions - but to be honest I just moved the check fro the if statement below out. I tend to change as little as possible to reduce the impact of anything going wrong. I suggest that this PR gets accepted and a subsequent cleanup commit would be made.

cziegeler: I agree

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
DominikSuess: The Felix Webconsole project is currently marked to be Java 6 compliant - this was just reproducing a pattern used already before for bsn - I just changed this based on suggestion by Carsten below to make sure to use a single method to obtain bsn & version

webconsole/src/main/java/org/apache/felix/webconsole/internal/core/BundlesServlet.java
DominikSuess: @cziegeler  please review if adjustments match your expectations

PR #5, opened 2020-03-03
FELIX-6189 - Make sure jar/zip files are jailed to the destination di…

coheigea:
…rectory


jbonofre:
Not sure it's a good idea. This kind of change already breaks Felix Utils. Let me do a pass and also reopen about Felix Utils.

karlpauls:
@jbonofre, did you manage to try it out?

jbonofre:
@karlpauls I will, but definitely, the same change broke Felix Utils (it has to be fixed by the way). So, I will do a pass on both. Sorry for the delay.

coheigea:
@jbonofre What's the problem with Felix Utils?

cziegeler:
five years without any activity, closing this PR

PR #6, opened 2020-03-03
Updating Commons Compress, FileUpload and IO

coheigea:


jbonofre:
LGTM. However, let me do a complete build/use pass just in case.

cziegeler:
Thanks for the PR!

PR #9, opened 2020-03-09
Convert to Interface without methods - FELIX-6238

stbischof:
https://issues.apache.org/jira/browse/FELIX-6238


bosschaert:
LGTM

@timothyjward WDYT?

timothyjward:
This method has always been a bit odd to read - it still seems odd to check that the methods returned by the interface are defined by the interface. 

As for the change, it looks correct, but would it be better to assign the output of `getMethods` to a local variable before the for loop so we don’t need to get the array twice?

bosschaert:
@stbischof would you have time to implement the change suggested by @timothyjward ?

stbischof:
@bosschaert sure. tomorrow!

stbischof:
@bosschaert here is the small change.

Do you have any ideas ho to handle this?:
https://issues.apache.org/jira/browse/FELIX-6239

PR #10, opened 2020-03-18
[converter] handle default methods - FELIX-6239

stbischof:
https://issues.apache.org/jira/browse/FELIX-6239

Default methods in interfaces exist since Java 8.
I did not found ONE way to invoke them.
I check the Java-Version and handles 2 different ways.

Maybe i should add an try-catch block do handle exceptions?

### DIFF
--- a/converter/converter/src/main/java/org/osgi/util/converter/ConvertingImpl.java
+++ b/converter/converter/src/main/java/org/osgi/util/converter/ConvertingImpl.java
@@ -17,6 +17,9 @@
 package org.osgi.util.converter;
 
 import java.lang.annotation.Annotation;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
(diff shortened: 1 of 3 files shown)


converter/converter/src/main/java/org/osgi/util/converter/ConvertingImpl.java
```diff
@@ -810,6 +813,32 @@ public Object invoke(Object proxy, Method method, Object[] args)
 					if (cls.isAnnotation()) {
 						val = method.getDefaultValue();
 					}
+					else if (method.isDefault())
+					{
+						double javaVersion = Double.parseDouble(
+							System.getProperty("java.class.version"));
+						double java8 = 52.0;
+						if (javaVersion > java8)
```
bosschaert: So invoking default methods is different for Java 8 and Java 9+?

bosschaert:
The PR looks good to me, I just find it surprising that there seems to be special code for just Java 8... Can the Java 8 code not use the MethodHandles approach?

converter/converter/src/main/java/org/osgi/util/converter/ConvertingImpl.java
stbischof: Yes. I didn't found one way that works in 8 and 9+

stbischof:
Each way works only with one. 8 oder 9+. 

The most information i found about this topic are there.

https://stackoverflow.com/questions/22614746/how-do-i-invoke-java-8-default-methods-reflectively/23990827

bosschaert:
Thanks @stbischof  I've merged your change.

bosschaert:
@stbischof is there more work needed or can https://issues.apache.org/jira/browse/FELIX-6239 be marked as resolved?

PR #11, opened 2020-03-22
[FELIX-6246] Packaged External Library Packages inside referenced Health Check Bundles

amitjoy:
For small IoT devices, it would definitely matter if someone needs to introduce Apache Commons Lang (~500KB) to use Felix Health Checks even if none of the other runtime bundles don't use Apache Commons. Therefore, this MR packages all external library packages inside the referenced bundles.

### DIFF
--- a/healthcheck/api/bnd.bnd
+++ b/healthcheck/api/bnd.bnd
@@ -1,12 +1,6 @@
-Bundle-Category: healthcheck
-
-Bundle-Description: ${project.description}
-
-Bundle-DocURL: https://felix.apache.org
-
-Bundle-License: Apache License, Version 2.0
-
-Bundle-Vendor: The Apache Software Foundation
-
-Export-Package: org.apache.felix.hc.api.*
-
+Bundle-Category     : healthcheck
+Bundle-Description  : ${project.description}
+Bundle-DocURL       : https://felix.apache.org
+Bundle-License      : Apache License, Version 2.0
+Bundle-Vendor       : The Apache Software Foundation
+Export-Package      : org.apache.felix.hc.api.*
--- a/healthcheck/core/bnd.bnd
+++ b/healthcheck/core/bnd.bnd
@@ -1,17 +1,8 @@
-Bundle-Category: healthcheck
-
-Bundle-Description: ${project.description}
-
-Bundle-DocURL: https://felix.apache.org
-
-Bundle-License: Apache License, Version 2.0
-
-Bundle-Vendor: The Apache Software Foundation
-
-Import-Package: org.quartz*;resolution:="optional", *
-
-Conditional-Package: org.apache.felix.utils.*
-
--removeheaders:\
-  Include-Resource,\
-  Private-Package
+Bundle-Category       : healthcheck
+Bundle-Description    : ${project.description}
+Bundle-DocURL         : https://felix.apache.org
+Bundle-License        : Apache License, Version 2.0
+Bundle-Vendor         : The Apache Software Foundation
+Import-Package        : org.quartz*;resolution:="optional", *
+-conditionalpackage   : org.apache.felix.utils.*, org.apache.commons.lang3.*
+-removeheaders        : Include-Resource, Private-Package
--- a/healthcheck/generalchecks/bnd.bnd
+++ b/healthcheck/generalchecks/bnd.bnd
@@ -1,15 +1,11 @@
-Bundle-Category: healthcheck
-
-Bundle-Description: ${project.description}
-
-Bundle-DocURL: https://felix.apache.org
-
-Bundle-License: Apache License, Version 2.0
-
-Bundle-Vendor: The Apache Software Foundation
-
-Conditional-Package: org.apache.commons.cli.*,org.apache.felix.utils.*
-
-Export-Package: org.apache.felix.hc.generalchecks.util
-
-Import-Package: org.apache.felix.rootcause*;resolution:="optional", *
+Bundle-Category     : healthcheck
+Bundle-Description  : ${project.description}
+Bundle-DocURL       : https://felix.apache.org
+Bundle-License      : Apache License, Version 2.0
+Bundle-Vendor       : The Apache Software Foundation
+Export-Package      : org.apache.felix.hc.generalchecks.util
+Import-Package      : org.apache.felix.rootcause*;resolution:="optional", *
+-conditionalpackage : \
+	org.apache.felix.utils.*,\
+	org.apache.commons.cli.*,\
+	org.apache.commons.lang3.*
--- a/healthcheck/webconsoleplugin/bnd.bnd
+++ b/healthcheck/webconsoleplugin/bnd.bnd
@@ -1,10 +1,6 @@
-Bundle-Category: healthcheck
-
-Bundle-Description: ${project.description}
-
-Bundle-DocURL: https://felix.apache.org
-
-Bundle-License: Apache License, Version 2.0
-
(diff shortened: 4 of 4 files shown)


healthcheck/api/bnd.bnd
```diff
@@ -1,12 +1,6 @@
...
-Bundle-Vendor: The Apache Software Foundation
-
-Export-Package: org.apache.felix.hc.api.*
-
+Bundle-Category     : healthcheck
+Bundle-Description  : ${project.description}
+Bundle-DocURL       : https://felix.apache.org
+Bundle-License      : Apache License, Version 2.0
+Bundle-Vendor       : The Apache Software Foundation
+Export-Package      : org.apache.felix.hc.api.*
```
ghenzler: No change required in api (commons-lang is not used)

healthcheck/core/bnd.bnd
```diff
@@ -1,17 +1,8 @@
...
--removeheaders:\
-  Include-Resource,\
-  Private-Package
+Bundle-Category       : healthcheck
+Bundle-Description    : ${project.description}
+Bundle-DocURL         : https://felix.apache.org
+Bundle-License        : Apache License, Version 2.0
+Bundle-Vendor         : The Apache Software Foundation
+Import-Package        : org.quartz*;resolution:="optional", *
+-conditionalpackage   : org.apache.felix.utils.*, org.apache.commons.lang3.*
```
ghenzler: Checked a build with and without conditional package - it looks like the whole commons lang library is pulled in ATM and hence the size of the hc core bundle is just the regular jar + the 435KB from commons lang ending up at a similar memory footprint. @amitjoy Can you do some tests on how to really get the file size down? (the optional include only taking what's needed works for org.apache.felix.utils.*, not sure why it does not for org.apache.commons.lang3.*)

healthcheck/generalchecks/bnd.bnd
```diff
@@ -1,15 +1,11 @@
...
+Bundle-Description  : ${project.description}
+Bundle-DocURL       : https://felix.apache.org
+Bundle-License      : Apache License, Version 2.0
+Bundle-Vendor       : The Apache Software Foundation
+Export-Package      : org.apache.felix.hc.generalchecks.util
+Import-Package      : org.apache.felix.rootcause*;resolution:="optional", *
+-conditionalpackage : \
+	org.apache.felix.utils.*,\
+	org.apache.commons.cli.*,\
+	org.apache.commons.lang3.*
```
ghenzler: @amitjoy If you use this bundle, embedding org.apache.commons.lang3 in both core and here could in total already result in a larger memory footprint (obviously depends on how much you actually use of the bundle)

healthcheck/webconsoleplugin/bnd.bnd
```diff
@@ -1,10 +1,6 @@
...
-Bundle-License: Apache License, Version 2.0
-
-Bundle-Vendor: The Apache Software Foundation
-
+Bundle-Category      : healthcheck
+Bundle-Description   : ${project.description}
+Bundle-DocURL        : https://felix.apache.org
+Bundle-License       : Apache License, Version 2.0
+Bundle-Vendor        : The Apache Software Foundation
+-conditionalpackage  : org.apache.commons.lang3.*
```
ghenzler: For an embedded use case I assume the webconsole is not relevant - I would not make changes here if not needed.

healthcheck/api/bnd.bnd
amitjoy: I just wanted to make the bnd files look alike. There is no real change in the API bundle.

healthcheck/core/bnd.bnd
amitjoy: I have also tested it yesterday and have noticed this discrepancy. The internal references to other different packages lead to the inclusion of several packages of org.apache.commons.lang3 which results to extra ~250KB in respective bundles. For the time being, I would propose to leave it as it is now. I will, in the meanwhile, think about it how we can trim down the references. We can definitely reimplement the referenced static factories from Apache Commons Lang3 once again that I tried too. But I feel it's rather reinventing the wheel. I will think about some slim-down approaches and keep you updated.

healthcheck/webconsoleplugin/bnd.bnd
amitjoy: Actually an embedded use case comprises both production-ready and developer-friendly gateways. I agree with you that such a web console bundle is not relevant for a production ready gateway but it is definitely required for developers in the development phase.

healthcheck/generalchecks/bnd.bnd
amitjoy: Yeah, I agree completely with you. This results in extra ~250KB in the respective bundles. I will think about some approaches that might help us trim down the file sizes.

healthcheck/api/bnd.bnd
ghenzler: Best to keep formatting changes out of PRs (because they are often based on personal preferences)

healthcheck/webconsoleplugin/bnd.bnd
ghenzler: So for server applications we do use the webconsole in production, I just thought that for `small IoT devices` you wouldn't use it at all (and maybe rather the gogo shell). But if you use it that's fine obviously, it just makes it one bundle more in your setup and the use case for embedding even weaker.

cschneider:
I am not in favour of embedding things like commons lang. It causes some additional problems:
1. If every bundles embeds some parts of other libraries the effective size of the application even grows.
2. If the embedding is not shaded (using different package names) then it is difficult to use such bundles with felix connect or atomos.
Instead of embedding I wonder if we can get rid of the depdendency

amitjoy:
@cschneider I agree completely with you and that's why we thought of delving more into it to either get rid of Apache Commons or replace it with custom implementations.

rotty3000:
I personally feel commons-lang has become obsolete since Java 8. I would strive to eliminate it's use wherever possible. Most times you use one or two functions and those are trivially implemented in Java 8.

ghenzler:
I'm also in favour of getting rid of the dependency altogether rather than trying to embed it. Unfortunately it's not just `StringUtils` that we use (that is usually quite quick to replace) but also classes like `StringEscapeUtils` and `WordUtils`. It will take a bit of effort to get rid of this (I'm not sure if copying  over some methods manually and just reference the source is ok license-wise). 

Another option is to make the dependency to make the bundle dependency to commons-lang optional and turn some features off if those classes are not there (e.g. the verbose txt rendering that has the dependency to `WordUtils`)

amitjoy:
@ghenzler @rotty3000 I agree with you all. It makes sense to get rid of the dependency. But it isn't very straight forward since I tried to do so. I felt like I am reinventing the wheel once again when it is already available in Lang 3. Some of the features that Felix HC uses can simply be done using simple Java 8 features but not all such as `WordUtils` as pointed out by @ghenzler. I am working on other parts of Felix HC to optimise as much as possible. And that's why I kept it open so that I don't forget to fix it as well. I have some plans to get rid of Apache Commons Lang dependency completely and write only the required methods from `WordUtils` from scratch.

rombert:
Perhaps it's worthwhile to remove all usages of commons-lang3 that are "trivial" and replaceable with Java 8, and then try using `Conditional-Package` again?

ghenzler:
I did some tests and we are able to remove commons lang completely. It's a bit of work but makes the most sense here. 

Closing this PR

PR #13, opened 2020-03-28
[HealthCheck] Fixed Warnings and Introduced Minor Code Refactorings

amitjoy:
In General Checks and Web Console Bundles


amitjoy:
@ghenzler Would you kindly have a look into it?

PR #16, opened 2020-04-09
[FELIX-6058] Re-added Bundle-Activator to gogo.jline bundle

glimmerveen:
Currently the gogo.jline bundle stand-alone is not usable. There was already an open PR on the felix repo (https://github.com/apache/felix/pull/184), but given the state of the felix repository, this will not get merged.

This PR ensures that the Bundle-Activator headers is set as instruction on the maven-bundle-plugin, making the gogo.jline bundle both usable stand-alone and in the context of Karaf's shell.core where the gogo.jline package is inlined.

### DIFF
--- a/gogo/jline/pom.xml
+++ b/gogo/jline/pom.xml
@@ -88,6 +88,7 @@
                 <artifactId>maven-bundle-plugin</artifactId>
                 <configuration>
                     <instructions>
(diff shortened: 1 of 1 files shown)


glimmerveen:
@jbonofre could you have a look at this PR? I believe it would both allow the current repackaging in Karaf and allow the bundle to be used as is, thus increasing the usability of this bundle.

jbonofre:
I will do the change in Karaf. +1 for the merge (already done).

PR #17, opened 2020-04-12
FELIX-6203 skip adding comments in pom.properties

doggy-dev:


doggy-dev:
Hi,
is everything ok with the pull request? Should I change something?

PR #19, opened 2020-04-21
[FELIX-6161] Replace local filtering of references in favour of relying on filtered ServiceListener callbacks

glimmerveen:
This PR contains the patch already attached to FELIX-6161, but now applied on the current state of the felix-dev repository.

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
@@ -172,8 +172,10 @@ public class BundleComponentActivator implements ComponentActivator
             ListenerInfo listenerInfo = listenerMap.get( serviceFilterString );
             if ( listenerInfo != null )
             {
-                listenerMap.remove( serviceFilterString );
-                m_context.removeServiceListener( listenerInfo );
+                if (listenerInfo.remove(listener)) {
+                    listenerMap.remove(serviceFilterString);
+                    m_context.removeServiceListener(listenerInfo);
+                }
             }
         }
     }
--- a/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
@@ -51,7 +51,6 @@ import org.apache.felix.scr.impl.metadata.ComponentMetadata;
 import org.apache.felix.scr.impl.xml.XmlHandler;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
-import org.osgi.framework.Filter;
 import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.ServiceEvent;
 import org.osgi.framework.ServiceListener;
@@ -103,7 +102,6 @@ public class BundleComponentActivator implements ComponentActivator
         @Override
         public void serviceChanged(ServiceEvent event)
         {
-            ServiceReference<?> ref = event.getServiceReference();
             ExtendedServiceEvent extEvent = new ExtendedServiceEvent(event);
             List<ExtendedServiceListener<ExtendedServiceEvent>> listeners;
             synchronized ( this )
--- a/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
@@ -98,108 +98,60 @@ public class BundleComponentActivator implements ComponentActivator
 
     private static class ListenerInfo implements ServiceListener
     {
-        private Map<Filter, List<ExtendedServiceListener<ExtendedServiceEvent>>> filterMap = new HashMap<>();
+        List<ExtendedServiceListener<ExtendedServiceEvent>> listeners = new ArrayList<>();
 
         @Override
         public void serviceChanged(ServiceEvent event)
         {
             ServiceReference<?> ref = event.getServiceReference();
-            ExtendedServiceEvent extEvent = null;
-            ExtendedServiceEvent endMatchEvent = null;
-            Map<Filter, List<ExtendedServiceListener<ExtendedServiceEvent>>> filterMap;
+            ExtendedServiceEvent extEvent = new ExtendedServiceEvent(event);
+            List<ExtendedServiceListener<ExtendedServiceEvent>> listeners;
             synchronized ( this )
             {
-                filterMap = this.filterMap;
+                listeners = this.listeners;
             }
-            for ( Map.Entry<Filter, List<ExtendedServiceListener<ExtendedServiceEvent>>> entry : filterMap.entrySet() )
+
+            for ( ExtendedServiceListener<ExtendedServiceEvent> forwardTo : listeners)
             {
-                Filter filter = entry.getKey();
-                if ( filter == null || filter.match( ref ) )
-                {
-                    if ( extEvent == null )
-                    {
-                        extEvent = new ExtendedServiceEvent( event );
-                    }
-                    for ( ExtendedServiceListener<ExtendedServiceEvent> forwardTo : entry.getValue() )
-                    {
-                        forwardTo.serviceChanged( extEvent );
-                    }
-                }
-                else if ( event.getType() == ServiceEvent.MODIFIED )
-                {
-                    if ( endMatchEvent == null )
-                    {
-                        endMatchEvent = new ExtendedServiceEvent( ServiceEvent.MODIFIED_ENDMATCH, ref );
-                    }
-                    for ( ExtendedServiceListener<ExtendedServiceEvent> forwardTo : entry.getValue() )
-                    {
-                        forwardTo.serviceChanged( endMatchEvent );
-                    }
-                }
+                forwardTo.serviceChanged( extEvent );
             }
+
             if ( extEvent != null )
             {
                 extEvent.activateManagers();
             }
-            if ( endMatchEvent != null )
-            {
-                endMatchEvent.activateManagers();
-            }
         }
 
-        public synchronized void add(Filter filter, ExtendedServiceListener<ExtendedServiceEvent> listener)
+        public synchronized void add(ExtendedServiceListener<ExtendedServiceEvent> listener)
         {
-            filterMap = new HashMap<>( filterMap );
-            List<ExtendedServiceListener<ExtendedServiceEvent>> listeners = filterMap.get( filter );
-            if ( listeners == null )
-            {
-                listeners = Collections.<ExtendedServiceListener<ExtendedServiceEvent>> singletonList( listener );
-            }
-            else
-            {
-                listeners = new ArrayList<>( listeners );
-                listeners.add( listener );
-            }
-            filterMap.put( filter, listeners );
+            listeners = new ArrayList<>(listeners);
+            listeners.add(listener);
         }
 
-        public synchronized boolean remove(Filter filter, ExtendedServiceListener<ExtendedServiceEvent> listener)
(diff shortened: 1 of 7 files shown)


scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
```diff
@@ -98,133 +98,82 @@
 
     private static class ListenerInfo implements ServiceListener
     {
-        private Map<Filter, List<ExtendedServiceListener<ExtendedServiceEvent>>> filterMap = new HashMap<>();
+        List<ExtendedServiceListener<ExtendedServiceEvent>> listeners = new ArrayList<>();
```
tjwatson, line 101: Shouldn't this just be a CopyOnWriteArrayList?

glimmerveen: It has been a while I made these changes, so I am not too sure, but I believe I mimicked the current approach, but with a List rather than a Map. Looking at the places where the list is modified, it looks similar to what a CopyOnWriteArrayList would do.

I don't mind changing it, though I am not familiar with possible additional reasons why it was done in this way.

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
```diff
@@ -98,133 +98,82 @@
...
-            ListenerInfo listenerInfo = listenerMap.get( className );
+            ListenerInfo listenerInfo = listenerMap.get( serviceFilterString );
             if ( listenerInfo != null )
             {
-                if ( listenerInfo.remove( filter, listener ) )
-                {
-                    listenerMap.remove( className );
-                    m_context.removeServiceListener( listenerInfo );
-                }
+                listenerMap.remove( serviceFilterString );
```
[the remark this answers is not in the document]
glimmerveen: You are right, this is not correct. I think it is missing the removal of the ExtendedServiceListener from the ListenerInfo first, and only removing it from the map if the ListenerInfo is empty.

glimmerveen:
> I have some requested changes. I also would like to see some tests that verify that the listener hooks now have the expected information.

Thanks for your feedback, I will also work on adding some tests to verify the behaviour.

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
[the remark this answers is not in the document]
tjwatson: Part of me that is concerned about why the code was the way it is currently.  I have to admit that I don't know the history of why though.  For example, I'm unsure what condition causes the ListenerInfo to even have multiple ExtendedServiceListeners in its list of listeners.  There appears to be a one-to-one mapping between a ServiceTracker and a DependencyManager and a DependencyManager manages a single reference and hence a single target filter for the reference.  I guess I don't see how multiple targets would come from a single DependencyManager.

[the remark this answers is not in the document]
glimmerveen: Perhaps it is related to current approach of effectively aggregating multiple references into a single ServiceListener. As this PR aims to remove this aggregation, perhaps this is something that can be further simplified?

djencks:
Trying to remember the distant past….

I think that the problem this whole thing was designed to solve was that it’s possible to have several references to the same service, even from the same component.  (I think this also solves it for multiple components with the same dependency, in the same bundle). If you register several listeners, then you get several events and unfortunate things can happen in between those events.  My code was an attempt to get only one event per dependency, no matter how many references that is spread over.  IIRC Arnoud’s change change the mapping from “all dependencies on some service”, where DS manages the filters, to “all dependencies with a certain filter on some service”, letting the service registry manage the filters.  This introduces some circumstances where multiple events might be needed, but still eliminates the most likely, where for instance you get the service in one place and the service property map in another.

I think my idea was to, for instance, with a static reference, get all the references to a particular service updated in between the deactivation and reactivation.  With entirely separate events there’s going to be a peculiar short lived state where the two references to the same service/filter will be referring to different instances.

David Jencks

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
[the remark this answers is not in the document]
djencks: That is my recollection of how it worked.  I'd recommend in this change aggregating all the references to a service with a particular filter from a particular bundle to get the minimum number of events.  Is that consistent with the "service on request"  you are trying to implement?

scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java
```diff
@@ -2283,8 +2283,8 @@ else if ( m_dependencyMetadata.getScope() == ReferenceScope.prototype_required )
             null, getName(), initialActive, refMap, classFilterString, eventFilter,
                     initialReferenceFilterString );
         ServiceTracker<T, RefPair<S, T>, ExtendedServiceEvent> tracker = new ServiceTracker<>(
-            bundleContext, m_customizer, initialActive, m_componentManager.getActivator(), eventFilter,
```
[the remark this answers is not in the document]
glimmerveen: Good catch, will remove the creation of the eventFilter as well.

glimmerveen:
I have updated the PR to include an integration test as well that validates two aspects of the use of ServiceListener discussed: 

1. Target filter properties visible to ListenerHook
2. If the same service is referenced multiple times from the same bundle, only a single ServiceListener is opened and maintained.

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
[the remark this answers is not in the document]
glimmerveen: @djencks I believe my current changes reflect the behaviour you suggested, and I have added a test to verify this.
The important change also here that supports my use-case is the fact that the target filter is now also visible to the ListenerHook's, allowing such an implementation to register a service that will match the target filter.

scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java
```diff
@@ -2239,52 +2238,16 @@ else if ( m_dependencyMetadata.getScope() == ReferenceScope.prototype_required )
         m_componentManager.getLogger().log(LogService.LOG_DEBUG, "Setting target property for dependency {0} to {1}",
                 null, getName(), target );
         BundleContext bundleContext = m_componentManager.getBundleContext();
```
tjwatson, line 2240: There is an additional block above that is no longer needed also:
```
        // eventFilter
        String eventFilterString;
        if (m_target != null
            && m_dependencyMetadata.getScope() == ReferenceScope.prototype_required)
        {
            final StringBuilder sb = new StringBuilder("(&").append(PROTOTYPE_SCOPE_CLAUSE).append(m_target).append(")");
            eventFilterString = sb.toString();
        }
        else if ( m_dependencyMetadata.getScope() == ReferenceScope.prototype_required )
        {
            eventFilterString = PROTOTYPE_SCOPE_CLAUSE;
        }
        else
        {
            eventFilterString = m_target;
        }
```

scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java
```diff
@@ -2239,52 +2238,16 @@ else if ( m_dependencyMetadata.getScope() == ReferenceScope.prototype_required )
...
-        Filter eventFilter = null;
```
tjwatson, line 2242: I tried to find if there is a specified behavior in the DS spec if the target has an invalid filter.  I could not find anything.  Regardless, the current Felix SCR impl seems to log an error here and then create this target `(component.id=-1)` filter to be used instead.

I suggest we still need this check with something like this:
```
        if (target != null)
        {
            try
            {
                FrameworkUtil.createFilter(target);
            }
            catch (InvalidSyntaxException e)
            {
                m_componentManager.getLogger().log(LogService.LOG_ERROR,
                    "Invalid syntax in target property for dependency {0} to {1}", null,
                    getName(), target);

                //create a filter that will never be satisfied  
                target = "(component.id=-1)";
            }
        }
        m_target = target;
```
Where we do the check on the incoming `target` param before setting it to the instance `m_target`

scr/src/test/resources/integration_test_FELIX_6161.xml
```diff
@@ -0,0 +1,64 @@
...
+        <reference
+                name="one"
+                interface="org.apache.felix.scr.integration.components.SimpleService"
+                cardinality="0..1"
+                policy="dynamic"
+                bind="setSimpleService"
+                unbind="unsetSimpleService"
+                target="(value=foo)"
+        />
+    </scr:component>
```
tjwatson, line 63: I suggest a third component that uses a different target filter from the other two.  And a forth that has an invalid filter for target.

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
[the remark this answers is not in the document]
tjwatson: OK, I have a better understanding of how the listeners are managed now!  I was getting stuck on the fact that we have a ServiceTracker created per reference, even when there are multiple references from the same bundle with the idententical target.  I was missing the fact that the ServiceTracker uses the `ExtendedServiceListenerContext` to manage the registration of the listener and that there is a single `ExtendedServiceListenerContext` per bundle.

scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
```diff
@@ -98,132 +98,83 @@
...
             {
                 extEvent.activateManagers();
             }
-            if ( endMatchEvent != null )
-            {
-                endMatchEvent.activateManagers();
-            }
         }
 
-        public synchronized void add(Filter filter, ExtendedServiceListener<ExtendedServiceEvent> listener)
```
tjwatson, line 150: Removal of the Filter param leaves an unused import above.

scr/src/main/java/org/apache/felix/scr/impl/manager/ServiceTracker.java
```diff
@@ -76,13 +76,6 @@
 	 * The Bundle Context used by this {@code ServiceTracker}.
 	 */
 	protected final BundleContext			context;
-	/**
-	 * The Filter used by this {@code ServiceTracker} which specifies the search
-	 * criteria for the services to track.
-	 * 
-	 * @since 1.1
-	 */
-	protected final Filter					eventFilter;
```
tjwatson, line 85: Leaves an unused import above.

scr/src/test/java/org/apache/felix/scr/impl/manager/SingleComponentManagerTest.java
```diff
@@ -51,15 +51,15 @@
     private ComponentActivator componentActivator = new ComponentActivator() {
 
         @Override
-        public void addServiceListener(String className, Filter filter,
```
tjwatson, line 54: Causes unused import

scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java
glimmerveen: I implemented this, but opted to use a different target in case of an invalid one in the component definition, that is guaranteed to be never be satisfied (inspiration drawn from the required capability in the osgi.cmpn API bundle).

tjwatson: Good idea, looks good.

PR #21, opened 2020-05-13
[FELIX-6274] Add a test demonstrating the LogService ClassCastException

timothyjward:
Create a simple test which shows component bundles using different incompatible versions of the log service
can't be properly handled by Felix SCR due to a ClassCastException

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/logger/LogServiceSupport.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/logger/LogServiceSupport.java
@@ -30,7 +30,7 @@ class LogServiceSupport
 {
     private final boolean r7Enabled;
 
-    private final LogService logService;
+    private final Object logService;
 
     private final Bundle bundle;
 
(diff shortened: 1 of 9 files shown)


tjwatson:
For your scenario is R6 or R7 log specification being used? or maybe both at the same time?  

My opinion is if SCR itself is wired to the R7 log package then it must always use the new Logger interface to log all messages.  If SCR is logging messages related to a specific component then SCR must call the LoggerFactory.getLogger(Bundle bundle, String name, Class loggerType) method passing the bundle declaring the component to obtain the Logger.  This is described at https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e41804

In that case I believe SCR must use its own BundleContext to get the LoggerFactory service NOT the context of the extendee bundle.  I think it would be unfortunate to resort to reflection to call the Logger because SCR is very noisy with its debug logging.  That is optimized out by the log service implementation if debug is not enabled with LoggerAdmin, but now we may put a reflective call before that optimization can be used.

Unfortunately Felix SCR also tries to support the R6 LogService (perhaps we should rip that out?).  For R6 LogService SCR must use the BundleContext of the extendee.  The R6 Declarative Service spec states:  

> When SCR must log an error message to the Log Service, it must use a Log Service obtained using the component's Bundle Context so that the resulting Log Entry is associated with the component's bundle.

So in that case I see the need to fall back to reflection.

But first I think we should decide if we want to support R6 LogService at all in SCR.  My vote is to rip that out and only support R7 which allows us to only use the context of SCR itself to obtain the LoggerFactory therefore avoiding any possible cast exceptions.

rotty3000:
Is this the same issue as FELIX-6281?

tjwatson:
> Is this the same issue as FELIX-6281?

No, it is similar in the scenario, but fixing the log impl FELIX-6281 issue will not fix this SCR issue.

tjwatson:
I got delayed, but I do have a fix ready that avoids the reflection.  I would prefer we use that solution instead of introducing reflection here.  I'll get a PR going now.

rotty3000:
@timothyjward There are two underlying issues.
a) a bug in Felix LogService
b) a bug in Felix SCR

issue a) is fixed here: https://issues.apache.org/jira/browse/FELIX-6077
And I'm pretty sure @tjwatson is close to having a fix for SCR.

tjwatson:
I opened PR #27 which includes your original test commit (rebased to latest master)

tjwatson:
This testcase got merged as part of another PR

PR #22, opened 2020-05-29
[FELIX-6281] Potential class space issues for getLogger(Bundle ...)

rotty3000:


log/src/main/java/org/apache/felix/log/LogServiceImpl.java
```diff
@@ -126,8 +125,8 @@ public Logger getLogger(String name) {
 
     @Override
     public Logger getLogger(Class<?> clazz) {
-        LogService logService = getLogService(clazz);
-        return logService.getLogger(clazz.getName());
+        Bundle bundle = FrameworkUtil.getBundle(clazz);
```
tjwatson, line 128: The class here is only to get the name of the logger.  It is not intended to get a logger for another bundle.

log/src/main/java/org/apache/felix/log/LogServiceImpl.java
```diff
@@ -137,29 +136,13 @@ public Logger getLogger(Class<?> clazz) {
 
     @Override
     public <L extends Logger> L getLogger(Class<?> clazz, Class<L> loggerType) {
-        LogService logService = getLogService(clazz);
-        return logService.getLogger(clazz.getName(), loggerType);
+        Bundle bundle = FrameworkUtil.getBundle(clazz);
```
tjwatson, line 139: Same comment about the class param

log/src/main/java/org/apache/felix/log/LogServiceImpl.java
```diff
@@ -137,29 +136,13 @@ public Logger getLogger(Class<?> clazz) {
...
-        }
-        BundleContext bundleContext = bundle.getBundleContext();
-        ServiceReference<LogService> serviceReference = bundleContext.getServiceReference(LogService.class);
-        return bundleContext.getService(serviceReference);
-    }
-
-    private LogService getLogService(Class<?> clazz) {
-        Bundle bundle = FrameworkUtil.getBundle(clazz);
-        return getLogService(bundle);
+        return m_loggerAdminImpl.getLogger(bundle, name, loggerType);
```
tjwatson, line 145: I think this should work

log/src/main/java/org/apache/felix/log/LogServiceImpl.java
rotty3000: got it! fixing

log/src/main/java/org/apache/felix/log/LogServiceImpl.java
rotty3000: fixed

PR #25, opened 2020-06-03
[FELIX-6218] Remove kxml notice

rovarga:
FELIX-6218 removed kxml from packaging, but did not update DEPENDENCIES and LICENSE. This patch finishes that up.


cziegeler:
Thanks @rovarga

PR #29, opened 2020-06-30
[FELIX-6265] Introduced Embedded Cron Scheduler

amitjoy:
Fixed https://issues.apache.org/jira/browse/FELIX-6265

Signed-off-by: Amit Kumar Mondal <admin@amitinside.com>

### DIFF
--- a/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/AsyncHealthCheckExecutor.java
+++ b/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/AsyncHealthCheckExecutor.java
@@ -39,8 +39,7 @@ import org.apache.felix.hc.core.impl.executor.HealthCheckFuture.Callback;
 import org.apache.felix.hc.core.impl.executor.HealthCheckResultCache;
 import org.apache.felix.hc.core.impl.scheduling.AsyncIntervalJob;
 import org.apache.felix.hc.core.impl.scheduling.AsyncJob;
-import org.apache.felix.hc.core.impl.scheduling.AsyncQuartzCronJob;
-import org.apache.felix.hc.core.impl.scheduling.QuartzCronSchedulerProvider;
+import org.apache.felix.hc.core.impl.scheduling.cron.quartz.CronJobFactory;
 import org.apache.felix.hc.core.impl.util.HealthCheckFilter;
 import org.apache.felix.hc.core.impl.util.lang.StringUtils;
 import org.osgi.framework.BundleContext;
@@ -48,7 +47,6 @@ import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.ServiceEvent;
 import org.osgi.framework.ServiceListener;
 import org.osgi.framework.ServiceReference;
-import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
@@ -72,17 +70,15 @@ public class AsyncHealthCheckExecutor implements ServiceListener {
     HealthCheckExecutorThreadPool healthCheckExecutorThreadPool;
     
     @Reference
-    QuartzCronSchedulerProvider quartzCronSchedulerProvider;
+    CronJobFactory cronJobFactory;
 
     @Activate
-    protected final void activate(final ComponentContext componentContext) throws InvalidSyntaxException {
-        this.bundleContext = componentContext.getBundleContext();
+    protected final void activate(final BundleContext bundleContext) throws InvalidSyntaxException {
         this.bundleContext.addServiceListener(this, "(objectclass=" + HealthCheck.class.getName() + ")");
-
         int count = 0;
         HealthCheckFilter healthCheckFilter = new HealthCheckFilter(bundleContext);
-        final ServiceReference[] healthCheckReferences = healthCheckFilter.getHealthCheckServiceReferences(HealthCheckSelector.empty(), false);
-        for (ServiceReference serviceReference : healthCheckReferences) {
+        final ServiceReference<?>[] healthCheckReferences = healthCheckFilter.getHealthCheckServiceReferences(HealthCheckSelector.empty(), false);
+        for (ServiceReference<?> serviceReference : healthCheckReferences) {
             HealthCheckMetadata healthCheckMetadata = new HealthCheckMetadata(serviceReference);
             if (isAsync(healthCheckMetadata)) {
                 if (scheduleHealthCheck(healthCheckMetadata)) {
@@ -94,7 +90,7 @@ public class AsyncHealthCheckExecutor implements ServiceListener {
     }
 
     @Deactivate
-    protected final void deactivate(final ComponentContext componentContext) {
+    protected final void deactivate() {
         this.bundleContext.removeServiceListener(this);
         this.bundleContext = null;
 
@@ -110,7 +106,7 @@ public class AsyncHealthCheckExecutor implements ServiceListener {
             // already deactivated?
             return;
         }
-        ServiceReference serviceReference = event.getServiceReference();
+        ServiceReference<?> serviceReference = event.getServiceReference();
         final boolean isHealthCheck = serviceReference.isAssignableTo(bundleContext.getBundle(), HealthCheck.class.getName());
 
         if (isHealthCheck) {
@@ -132,22 +128,13 @@ public class AsyncHealthCheckExecutor implements ServiceListener {
     }
 
     private boolean scheduleHealthCheck(HealthCheckMetadata descriptor) {
-
         try {
             AsyncJob healthCheckAsyncJob = null;
-            
             if (isAsyncCron(descriptor)) {
-            
-                try {
-                    healthCheckAsyncJob = new AsyncQuartzCronJob(getAsyncJob(descriptor), quartzCronSchedulerProvider, "job-hc-" + descriptor.getServiceId(), "async-healthchecks", descriptor.getAsyncCronExpression());
-                } catch(ClassNotFoundException|NoClassDefFoundError e) {
-                    LOG.warn("Can not schedule async health check '{}' with cron expression '{}' since quartz library is not on classpath", descriptor.getName(), descriptor.getAsyncCronExpression());
-                    return false;
-                }
+                healthCheckAsyncJob = cronJobFactory.createNewJob(getAsyncJob(descriptor), "job-hc-" + descriptor.getServiceId(), "async-healthchecks", descriptor.getAsyncCronExpression());
             } else if (isAsyncInterval(descriptor)) {
                 healthCheckAsyncJob = new AsyncIntervalJob(getAsyncJob(descriptor), healthCheckExecutorThreadPool, descriptor.getAsyncIntervalInSec());
             }
-
             if (healthCheckAsyncJob != null) {
                 healthCheckAsyncJob.schedule();
                 registeredJobs.put(descriptor, healthCheckAsyncJob);
@@ -155,12 +142,10 @@ public class AsyncHealthCheckExecutor implements ServiceListener {
             } else {
                 return false;
             }
-
         } catch (Exception e) {
             LOG.warn("Could not schedule job for " + descriptor + ". Exception: " + e, e);
             return false;
         }
-
     }
 
     private Runnable getAsyncJob(HealthCheckMetadata descriptor) {
--- a/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/HealthCheckExecutorImpl.java
+++ b/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/HealthCheckExecutorImpl.java
@@ -288,9 +288,9 @@ public class HealthCheckExecutorImpl implements ExtendedHealthCheckExecutor, Ser
 
 
     /** Create the health check meta data */
-    private List<HealthCheckMetadata> getHealthCheckMetadata(final ServiceReference... healthCheckReferences) {
+    private List<HealthCheckMetadata> getHealthCheckMetadata(final ServiceReference<?>... healthCheckReferences) {
         final List<HealthCheckMetadata> descriptors = new LinkedList<HealthCheckMetadata>();
-        for (final ServiceReference serviceReference : healthCheckReferences) {
+        for (final ServiceReference<?> serviceReference : healthCheckReferences) {
             final HealthCheckMetadata descriptor = getHealthCheckMetadata(serviceReference);
 
             descriptors.add(descriptor);
@@ -300,7 +300,7 @@ public class HealthCheckExecutorImpl implements ExtendedHealthCheckExecutor, Ser
     }
 
     /** Create the health check meta data */
-    private HealthCheckMetadata getHealthCheckMetadata(final ServiceReference healthCheckReference) {
+    private HealthCheckMetadata getHealthCheckMetadata(final ServiceReference<?> healthCheckReference) {
         final HealthCheckMetadata descriptor = new HealthCheckMetadata(healthCheckReference);
         return descriptor;
     }
--- a/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/HealthCheckExecutorThreadPool.java
+++ b/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/HealthCheckExecutorThreadPool.java
@@ -17,6 +17,7 @@
  */
 package org.apache.felix.hc.core.impl.executor;
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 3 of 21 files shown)


healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/AsyncHealthCheckExecutor.java
```diff
@@ -138,13 +135,7 @@ private boolean scheduleHealthCheck(HealthCheckMetadata descriptor) {
...
             
             if (isAsyncCron(descriptor)) {
-            	
-                try {
-    				healthCheckAsyncJob = new AsyncQuartzCronJob(getAsyncJob(descriptor), quartzCronSchedulerProvider, "job-hc-" + descriptor.getServiceId(), "async-healthchecks", descriptor.getAsyncCronExpression());
-                } catch(ClassNotFoundException|NoClassDefFoundError e) {
-                    LOG.warn("Can not schedule async health check '{}' with cron expression '{}' since quartz library is not on classpath", descriptor.getName(), descriptor.getAsyncCronExpression());
-                    return false;
-                }
+            	healthCheckAsyncJob = new AsyncCronJob(getAsyncJob(descriptor), healthCheckCronScheduler, "job-hc-" + descriptor.getServiceId(), descriptor.getAsyncCronExpression());
```
JochenHiller, line 138: Should be indented 2 spaces to right

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronScheduler.java
```diff
@@ -0,0 +1,170 @@
...
+import java.time.Duration;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
```
JochenHiller, line 14: Is it intended to use SLF4j logger und not an OSGi logger?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronScheduler.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Thread that makes all scheduling job.
+     */
+    public final SchedulerThread schedulerThread;
+
+    /**
+     * If scheduler is active or not.
+     */
+    private boolean isActive;
```
JochenHiller, line 44: Should `isActive` be `volatile` ? 

The access to it is not synchronized, so changes between `shutdown`, `isShutdown` and `run` might not be visible in other threads.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
```diff
@@ -0,0 +1,53 @@
+package org.apache.felix.hc.core.impl.executor.async.cron;
+
+import org.slf4j.Logger;
```
JochenHiller, line 3: Same here: SLF4J Logger vs. OSGi Logger?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
```diff
@@ -0,0 +1,53 @@
...
+            final CronParser sequenceGenerator) {
+        this.scheduler = scheduler;
+        this.cronJob = cronJob;
+        this.sequenceGenerator = sequenceGenerator;
+    }
+
+    @Override
+    public void run() {
+        try {
+        	cronJob.run();
```
JochenHiller, line 28: Indent should be 2 spaces to left

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
```diff
@@ -0,0 +1,53 @@
+package org.apache.felix.hc.core.impl.executor.async.cron;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class CronSchedulerTask implements Runnable {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public final CronJob cronJob;
```
JochenHiller, line 10: Why are these 3 variables public? Would `package` scope not be enough?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
```diff
@@ -0,0 +1,53 @@
+package org.apache.felix.hc.core.impl.executor.async.cron;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class CronSchedulerTask implements Runnable {
```
JochenHiller, line 6: Could be the whole class here be in `package` scope?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
```diff
@@ -0,0 +1,97 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.apache.felix.hc.core.impl.scheduling;
+
+import org.apache.felix.hc.core.impl.executor.async.cron.CronJob;
+import org.apache.felix.hc.core.impl.executor.async.cron.CronScheduler;
+import org.apache.felix.hc.core.impl.executor.async.cron.HealthCheckCronScheduler;
+import org.slf4j.Logger;
```
JochenHiller, line 23: Same here: SLF4J Logger vs. OSGi Logger?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ *  Runs health checks that are configured with a cron expression for asynchronous execution. 
+ */
+public class AsyncCronJob extends AsyncJob {
+	private final Logger logger = LoggerFactory.getLogger(getClass());
```
JochenHiller, line 30: Indent 2 spaces to left

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
```diff
@@ -0,0 +1,97 @@
...
+        try {
+            CronScheduler scheduler = healtchCheckScheduler.getScheduler();
+            scheduler.schedule(cronJob);
+            logger.info("Scheduled job {} with trigger {}", cronJob.name(), cronJob.cron());
+            return true;
+        } catch (Exception e) {
+            logger.error("Could not schedule job for " + runnable + ": " + e, e);
+            return false;
+        }
+
```
JochenHiller, line 51: Remove empty line

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
```diff
@@ -0,0 +1,97 @@
...
+    	
+    	private String name;
+    	private String cronExpression;
+    	
+        public HealthCheckJob(String cronExpression, String name) {
+        	this.name = name;
+        	this.cronExpression = cronExpression;
+		}
+        
+		@Override
```
JochenHiller, line 83: Indent to same column like method

healthcheck/core/src/test/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final Object SYNC_OBJECT = new Object();
+
+    @Test
+    public void checkSchedulerCheckSingleThread() throws InterruptedException {
+        try (final CronScheduler scheduler = new CronScheduler(Executors.newFixedThreadPool(1), 0, 1, TimeUnit.SECONDS, "T1")) {
+            final int nIterations = 3;
+
+            final SampleCronJob1 service = new SampleCronJob2(s -> {
+                System.out.println("Iteration: " + s.counter);
```
JochenHiller, line 21: Is it OK to write to console during tests in Apache Felix project?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/AsyncHealthCheckExecutor.java
amitjoy: There was a long discussion earlier to use specific formatting standard for Felix but as far as I got to know, it is not decided and that's why I wrote everything without any formatting applied. In Eclipse, the changes look good but due to formatting standard in GitHub, it looks different.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronScheduler.java
amitjoy: Since this class is not leveraging SCR, I used slf4j and slf4j is also used in many other places in Felix HC.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronScheduler.java
amitjoy: Fixed

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
amitjoy: Since this class is not leveraging SCR, I used slf4j

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
amitjoy: fixed

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
amitjoy: fixed

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
amitjoy: Since we don't have any proper formatting standard for Felix, I write everything without any formatting applied.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTask.java
amitjoy: In Eclipse, this looks good but due to formatting standard in GitHub, it looks different here.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
amitjoy: fixed

healthcheck/core/src/test/java/org/apache/felix/hc/core/impl/executor/async/cron/CronSchedulerTest.java
amitjoy: fixed

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
amitjoy: In Eclipse, this looks good but due to formatting standard in GitHub, it looks different here.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/scheduling/AsyncCronJob.java
amitjoy: Since this class is not leveraging SCR, I used slf4j instead.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/executor/async/AsyncHealthCheckExecutor.java
amitjoy: similarly here

ghenzler:
Thanks @amitjoy for providing  the PR! So I tried to build/install it and had a look at the code:

- I got the error message “Too many files with unapproved license” during build because the copyright header is missing 
- constructor injection in DS is a fairly recent feature, but with the HC module we want to be backwards compatible as much as possible. In HealthCheckCronScheduler, let’s use @Activate and @Reference as the other SCR components do
- IMHO it’s wise to keep quartz support in if the quartz bundle is installed since quartz supports more complex expressions like e.g. 0 15 10 ? * 6#3  (also potentially in the future there could be support for more exotic expression) - so org.apache.felix.hc.core.impl.scheduling.cron can just provide the base implementation (for the case quartz is not available), for complex expressions the users can always install quartz. Comparing to master, best case this PR would only rename AsyncQuartzCronJob to AsyncCronJob and choose the underlying implementation automatically
- the PR introduces quite a few tabs - please always use four spaces for indentation (to be in line with the other code)… (just found that QuartzCronSchedulerProvider.deactivate() also had tabs, which is my tech debt but still: Let’s use only spaces to be consistent)
- classes from org.apache.felix.hc.core.impl.executor.async.cron should be moved to org.apache.felix.hc.core.impl.scheduling.cron (since org.apache.felix.hc.core.impl.scheduling is used for async health check execution and for the HealthCheckMonitor, I refactored this when I introduced HealthCheckMonitor)
- I think the HealthCheckCronScheduler can directly run on HealthCheckExecutorThreadPool without creating a new ExecutorService (however then CronScheduler should not shut down that scheduler)

So this quite a bit of adjustments, let me know if you want to go for it (I can also continue on it if you prefer).

amitjoy:
@ghenzler Thanks a lot for your valuable feedback. I agree with you completely that it makes sense to support cron with and without the availability of Quartz. It will enable consumers to decide whether to use Quartz or the tiny cron scheduler that comes with it. Yeah, you are right that it is quite a bit of changes but I would love to amend the necessary changes. I will keep you updated on the progress.

amitjoy:
@ghenzler I am almost done with the changes. I would be able to update this PR by the end to today or early tomorrow.

ghenzler:
@amitjoy thanks for taking a look, I just created 8f78a180b24740a2c871 to make it easier for you because when looking into it I found the current code is a bit confusing to extend. So with 8f78a180b24740a2c871 you should be just able to extend on AsyncSimpleCronJob without touching much else... the cron specific code can go in package `org.apache.felix.hc.core.impl.scheduling.cron` that should be created.

amitjoy:
@ghenzler Don't worry about it. I have touched as little as possible and maintained a clean separation of packages for better comprehension. Thanks a lot again for your quick assistance.

amitjoy:
@ghenzler @JochenHiller I have updated the MR. You can now have a look.

amitjoy:
Just for the info: I have followed the proposal by @ghenzler to use Quartz optionally if Quartz packaged is wired to the bundle and if not, we will fallback to the usage of a tiny embedded cron scheduler. This will enable Felix HC to be used in small footprint IoT edge devices as well.

ghenzler:
Hi @amitjoy thanks for the contribution. Based on your PR I made the following improvements in da6061d2d76b5926b77b82:

* I remove a lot of boilerplate classes (to keep it as minimal as possible this will be good for IoT)
* I introduced support for 7th field year for at least `*`
** `*` could be easily supported because it is the same as omitting it, I had an existing expression that was failing
** I removed the shortcut `@reboot` (because for that we would need proper year field support, PR welcome)
* The class `EmbeddedCronParserTest` was using the deprecated Date constructor, fixed that and moved all tests for `EmbeddedCronParser` into `EmbeddedCronParserTest`
* The class `EmbeddedCronScheduler` is a lot simpler now and just uses the JDK `ScheduledExecutorService` for without complicated wait/notify mechanics. 

 Everything is together on master now, please test with your setup if everything works as expected.

PR #30, opened 2020-07-06
[FELIX-6295] Introduced Initial Delay for Asynchronous Interval

amitjoy:
https://issues.apache.org/jira/browse/FELIX-6295

### DIFF
--- a/healthcheck/api/src/main/java/org/apache/felix/hc/api/execution/HealthCheckMetadata.java
+++ b/healthcheck/api/src/main/java/org/apache/felix/hc/api/execution/HealthCheckMetadata.java
@@ -45,8 +45,9 @@ public class HealthCheckMetadata {
 
     private final String asyncCronExpression;
     private final Long asyncIntervalInSec;
+    private final Long asyncInitialDelayForIntervalInSec;
 
-    private final ServiceReference serviceReference;
+    private final ServiceReference<?> serviceReference;
 
     private final Long resultCacheTtlInMs;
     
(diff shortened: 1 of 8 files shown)


healthcheck/api/src/main/java/org/apache/felix/hc/api/execution/HealthCheckMetadata.java
```diff
@@ -130,6 +132,12 @@ public String getAsyncCronExpression() {
     public Long getAsyncIntervalInSec() {
         return asyncIntervalInSec;
     }
+    
+    /** Return the initial delay for interval in sec used for asynchronous execution.
+     * @return the initial delay for the async interval (or <code>null</code> if not set)  */
```
triller-telekom, line 137: Won't this be `0` instead of `null`, if it is not set? Because the default implementation sets it to `0`.

amitjoy: Yes, but the object returned here is `Long` which has been purposefully done to conform to the existing coding convention in Felix HC

ghenzler:
See https://issues.apache.org/jira/browse/FELIX-6295?focusedCommentId=17152074&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17152074 - I'm reluctant to introduce a new property if cron expressions can be used instead.

ghenzler:
See discussion in https://issues.apache.org/jira/browse/FELIX-6295 (closing this PR)

PR #35, opened 2020-07-16
FELIX-6279 Unchanged bundle does not obey force policy 

pkriens:
The force policy indicates that a bundle's configuration should
always override any local changes. 

However, the code contained an optimization that if the configuration
in a bundle was unchanged, it would ignore its configuration
completely, ignoring any force policies.

This patch overrides the optimization in ConfigList for records
where the policy is set to force.

- A test is added that failed on the current bundle and now passes
- The OSGi Test Suite has been run on the code, the XML file is
  included in the patch

See https://issues.apache.org/jira/browse/FELIX-6279?filter=-2


configurator/TEST-org.osgi.test.cases.configurator-7.0.0.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bosschaert, line 1: Isn't this file normally what is produced during a surefire test run. Do we need it as part of the sources?

pkriens: No, you need to run the OSGi test suite and that is restricted and not part of the build sadly. I created a bnd workspace and created a run configuration to run these test suites. I am not sure how this is normally handled but thought it best to include it as proof that it passes the OSGi CT.

PR #36, opened 2020-07-24
[scr] Logging Extension

pkriens:
The OSGi specifications mandates component relative
logging on the component's bundle and using the
implementation class as logger name.

This can be quite inconvenient since SCR is rather
enthusiastic logger. If the component's bundle has
DEBUG logging enabled, the customer's logging messages
tend to get overshadowed.

Also, the component logger could be disabled, but
then the bundle logger was used, and if the bundle
logger was disabled, the SCR logger was used. This
is quite inconvenient.

The code was not properly handling the case when
a LoggerFactory would disappear or a bundle would
stop.

Last, the code looked like it had grown a lot over time.

This patch will provide an (almost) source compatible
version of the SCR logging class ScrLogger, ComponentLogger,
and BundleLogger, only their creation differs.

The default behavior is (or should be) as today. There
is one exception, we will not delegate to the parent
since this seems a bug. The logger hierarchy is much better
handled by the Logger Admin.

The extension has a new configuration property:

	ds.log.extension ::= 'true' | 'false'

If set to true, the logging is not required to conform
to the OSGi specification.

ScrLogManager – Access point for SCR logging. It provides
a static method to create a ScrLogger. The ScrLogger gives
access to a BundleLogger, and this gives access to a
ComponentLogger.

The stdout logger has been removed and the fallback to 
System.out/err is now handled in the ScrLogManager.

ExtLogManager – Implements the extension

LogManager – However, the internals are quite different. First
there is a LogManager that handles the interaction 
between the OSGi service registry and the LoggerFactory,
cleaning up when the factory goes or the bundle goes.
This class has no _policy_. It uses a _facade_ to handle the
life cycle issues.

The InternalLogger had a Level that was used by all the code but the
API of the logger objects was defined by the AbstractLogger. Since
the InternalLogger had no meaning anymore, it is used as the
base interface for the logger _interfaces_. The scr, bundle, and 
component logger objects are now almost empty interfaces.

Package looks busier than I like, and many classes are almost empty
but this way there is almost no change in the rest of the code base.

## Testing

An extensive JUnit test was added in LoggerTest that has almost full
code coverage.


pkriens:
This PR belongs to https://issues.apache.org/jira/browse/FELIX-6293

amitjoy:
FYI I have tested these changes and it works perfectly. The newly introduced framework property works as a toggle to extend the SCR logging. I switched it on and off respectively to test the extension.

tjwatson:
It took me some time to start reviewing this.  But my first question is about how OSGi LoggerFactory objects are obtained.  With PR #27 I changed SCR to only use a single LoggerFactory using SCR's bundle context and then obtain Loggers on behalf of other bundles by using the method `org.osgi.service.log.LoggerFactory.getLogger(Bundle, String, Class<L>)`

Unfortunately my changes in PR #27 broke the test case that Tim Ward provided such that it was not really testing properly.  This allowed your changes to pass the test `org.apache.felix.scr.integration.Felix6274Test.test_incompatible_log_service_version()` when it should have gotten a cast exception.  I am going to merge PR #40 to fix the test, but that then makes your changes fail the test.

I think your changes would be more simple if you only had a single LoggerFactory obtained with the SCR context that you used to obtain all the OSGi Logger instances similar to how the code worked before.  There should be no need to have complicated tracking of each LoggerFactory registration from the perspective of the component bundles.

pkriens:
I can change this. I tried to change as close as possible to the code I found.

scr/src/main/java/org/apache/felix/scr/impl/logger/ScrLogManager.java
```diff
@@ -0,0 +1,293 @@
...
+					return logger.isDebugEnabled();
+				}
+			} else {
+				return getLogLevel().implies(level);
+			}
+		}
+
+		@Override
+		public void log(Level level, String format, Throwable ex, Object... arguments) {
+			log(level, format(format, arguments), ex);
```
tjwatson, line 151: The `format` work should not be done here when `isLogEnabled(level)` is false.

scr/src/main/java/org/apache/felix/scr/impl/logger/ScrLogManager.java
```diff
@@ -0,0 +1,293 @@
...
+		}
+
+		@Override
+		public void log(Level level, String message, Throwable ex) {
+			if (prefix != null && prefix.length() > 0) {
```
tjwatson, line 156: Can move this into the if check for `isLogEnabled(level)` to avoid unnecessary work.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+ * A class that tracks the LoggerFactory of a bundle and its associated Logger objects. 
+ * When a bundle is stopped and/or the service is unregistered, the logger objects are cleaned.
+ * 
+ * The basic technique is to use a facade. Instead of returning a log object, we return a facade. The 
+ * methods delegate to the actual logger object. If there is no logger object, we create one.
+ * 
+ * The LogDomain represents every bundle. Per LogDomain, we keep the facades. If the factory goes,
+ * we reset the facades.
+ */
+class LogManager extends ServiceTracker<LoggerFactory, LoggerFactory> {
```
tjwatson, line 47: Can this class be folded into the default `ScrLogManager`?  It doesn't seem that anything else directly extends this class except `ScrLogManager` and then `ExtLogManager` extends `ScrLogManager`.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+
+	@Override
+	public void removedService(ServiceReference<LoggerFactory> reference, LoggerFactory service) {
+		if (!closed.get()) {
+			reset();
+		}
+	}
+
+	private void reset() {
+		for (LogDomain domain : domains.values()) {
```
tjwatson, line 82: Need `synchronized (domains)` around the for loop to avoid concurrent modification exception.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+		void reset() {
+			for (LoggerFacade facade : facades) {
+				facade.reset();
+			}
+		}
+
+		@SuppressWarnings("resource")
+		@Override
+		public void bundleChanged(BundleEvent event) {
+			if (event.getBundle() == bundle && event.getType() == BundleEvent.STOPPED && !closed.get()) {
```
tjwatson, line 124: We can do this with a single `BundleListener`.  I suggest `LogManager` implement `BundleListener` because this method is managing the map from `LogManager`

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+			return domain.bundle;
+		}
+
+		String getName() {
+			return name;
+		}
+
+	}
+
+	public void close() {
```
tjwatson, line 185: This does not call `super.close` from ServiceTracker?

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+				domain = new LogDomain(bundle);
+				domains.put(bundle, domain);
+			}
+			return domain;
+		}
+	}
+
+	@Override
+	public void removedService(ServiceReference<LoggerFactory> reference, LoggerFactory service) {
+		if (!closed.get()) {
```
tjwatson, line 76: This does not call `super.removedService`.  I don't see where the LoggerFactory service would be ungotten without this.  This also does not `null` out the `LogManager.factory` field.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+
+	<T> T getLogger(Bundle bundle, String name, Class<T> type) {
+		return type.cast(getLogDomain(bundle).getLogger(name));
+	}
+
+	LoggerFactory getLoggerFactory() {
+		int trackingCount = getTrackingCount();
+		if (this.trackingCount < trackingCount) {
+			this.trackingCount = trackingCount;
+			factory = getService();
```
tjwatson, line 95: In the case where a new higher ranked `LoggerFactory` has been registered don't we need to reset all the loggers from the previous `LoggerFactory`?

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
```diff
@@ -0,0 +1,203 @@
...
+		private final Set<LoggerFacade>	facades	= new HashSet<>();
+
+		LogDomain(Bundle bundle) {
+			this.bundle = bundle;
+			scrContext.addBundleListener(this);
+			open();
+		}
+
+		void reset() {
+			for (LoggerFacade facade : facades) {
```
tjwatson, line 116: Access to facades over this for loop needs to be thread safe with the `getLogger(String)` method to avoid co-modification errors.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
pkriens: I'd rather keep it separate for cohesion. The LogManager is the part that handles the life cycle and could theoretically be used in other projects. The ScrLogManager handles the interface to SCR. The life cycle is tricky as it is and combining it with the oddities of SCR makes it way more complicated than separating these parts.

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
pkriens: Since the single service tracker & the higher priority made the locking harder, I rewrote these parts. Good catch

scr/src/main/java/org/apache/felix/scr/impl/logger/LogManager.java
pkriens: fixed & tested

tjwatson:
I have some suggested fixes that I would like merged into Peter's branch pkriens/scr-logging before merging this PR:  https://github.com/pkriens/felix-dev/pull/1

PR #46, opened 2020-09-04
FELIX-6325 - OSGi log package is mandatory but imported as optional

tjwatson:
Make the log package optional again.
The implementation still will only use
an OSGi R7 log service implementation
if one is available.


jbonofre:
@tjwatson Awesome ! Thanks. I'm testing it !

jbonofre (approved):
I tested this change successfully with Karaf 4.2.10-SNAPSHOT. Thanks a lot !
I propose to merge the PR and move forward with SCR 2.1.24 release (else I can't update SCR version in Karaf 4.2.x).

PR #51, opened 2020-09-30
CQ-4305335 HTTP Session not invalidated over HTTPS

abhishekgarg18:


cziegeler:
A similar fix has already been applied

PR #52, opened 2020-09-30
[FELIX-6338] Prevent Null pointer exception in fileinstall watcher

mkep-dev:
An OVERFLOW event might have a context that is null. As a consequence the Path.resolve call will throw a null pointer exception

See https://issues.apache.org/jira/browse/FELIX-6338


rotty3000:
Better late than never... Thank you for the contribution.

PR #59, opened 2020-11-05
Introduce configuration options for setting felix shutdown and reload timeout

lucapalano:
JIRA REF: [FELIX-6356](https://issues.apache.org/jira/browse/FELIX-6356)

### DIFF
--- a/webconsole/src/main/java/org/apache/felix/webconsole/WebConsoleConstants.java
+++ b/webconsole/src/main/java/org/apache/felix/webconsole/WebConsoleConstants.java
@@ -18,6 +18,9 @@
  */
 package org.apache.felix.webconsole;
 
+import java.util.Dictionary;
+
+import org.apache.felix.webconsole.internal.servlet.OsgiManager;
 
 /**
  * WebConsoleConstants provides some common constants that are used by plugin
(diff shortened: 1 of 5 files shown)


cziegeler:
@lucapalano Thanks for your patch and sorry for the very late response.
The OsgiManager class has already support for configuration admin (through registering a ManagedService). I think your patch can be simplified, if it just leverages OsgiManager.getDefaultConfiguration

lucapalano:
Hi @cziegeler,

Thank you for your feedback about. I pushed new commits trying to simplify the previous one.

I'll look forward for your feedback.

Best regards,
Luca

cziegeler:
@lucapalano Thanks again, looks good, but :) by looking at your PR I realized that there is no good way to pass down the configuration from the OsgiManager to a plugin.
I think we could make this easier, if the OsgiManager would simply add the configuration as a request attribute (similar to the others here: https://github.com/apache/felix-dev/blob/master/webconsole/src/main/java/org/apache/felix/webconsole/internal/servlet/OsgiManager.java#L594) before calling the plugin. This would remove the need for all of the (static) managed service/configuration handling

lucapalano:
@cziegeler I see your point. I made it more simple by using, as you suggested, the "request" parameter. "shutdown" and "reload" timeouts attributes are set for the "VMstat" plugin only.

cziegeler:
@lucapalano Awesome, thank you - how about you always just set the full configuration map as an attribute for every plugin - that's not breaking any plugin. But then we don't have a hard dependency from the OsgiManager to the VMstat plugin - and the VMstat plugin just picks whatever it needs from the configuration map?

lucapalano:
@cziegeler now, shutdown and reload timeouts can be accessed via `OsgiManager.PROP_SHUTDOWN_TIMEOUT` and `OsgiManager.PROP_RELOAD_TIMEOUT` configuration props.

The configuration params can be always set, in the webconsole configuration file, by using the props:
- felix.webconsole.shutdown.timeout (OsgiManager.FRAMEWORK_SHUTDOWN_TIMEOUT);
- felix.webconsole.reload.timeout (OsgiManager.FRAMEWORK_RELOAD_TIMEOUT);

cziegeler (approved):

cziegeler:
@lucapalano Thanks for all the work and sorry for asking for all these changes.

PR #65, opened 2020-12-13
FELIX-6369: Implement R8 core spec

karlpauls:


karlpauls:
This passes the R8 core ct with java8, 11, 15, and 16-ea. Additionally, it passes the R7 ct with java8 and the results equal Felix 6.x for R6 (using java8). It doesn't require any `--add-opens` anymore and has the required native-image instructions to be used in a native image (together with atomos).

PR #67, opened 2020-12-15
FELIX-6163: do not use System.currentTimeMillis() to check timeouts

karlpauls:


framework/src/main/java/org/apache/felix/framework/util/ThreadGate.java
```diff
@@ -72,14 +74,14 @@ public synchronized void setMessage(Object msg)
     **/
     public synchronized boolean await(long timeout) throws InterruptedException
     {
-        long start = System.currentTimeMillis();
+        long start = TimeUnit.NANOSECONDS.toMillis(System.nanoTime());
```
tjwatson, line 77: Could convert the timeout input to nanoseconds instead of converting system.nanoTime to millis possibly multiple times.

bjhargrave:
https://github.com/osgi/osgi/pull/149/commits/cf7d482e956c89269fa1518d15391bd229408fa4 is the fix in the OSGi source repo.

PR #71, opened 2021-02-26
FELIX-6390 Refactor the default authentication mechanism of the webconsole to be a WebConsoleSecurityProvider2

enapps-enorman:
For [FELIX-6390](https://issues.apache.org/jira/browse/FELIX-6390)

The use case is that a webconsole plugin needs to make http requests outside of the OsgiManager servlet to retrieve some information to display in the plugin UI.   The goal is that the security checking of that other endpoint would perform the same security checks that would be needed to access the webconsole itself.  Reusing the WebConsoleSecurityProvider service in both places would be ideal.

The proposal is to refactor the default "basic" authentication mechanism of the webconsole into a WebConsoleSecurityProvider class and expose it as a service.  A very low service.ranking of this last resort security provider should ensure that any other WebConsoleSecurityProvider component that exists would be used instead.


enapps-enorman:
In the interest of moving forward, would it be possible to get a review and ruling on this proposal?

enapps-enorman:
Felix Committers:  I have not gotten any feedback on the proposal,  so I'm asking again for a review and ruling on the proposal.  If you want to say no then that would be better than just ignoring the proposal.  I really need a resolution one way or the other.

cziegeler (approved):
Lgtm, thanks

PR #73, opened 2021-04-06
adding test cases for CQ-4318856

abhishekgarg18:


cziegeler:
The problem was not in the converter but a concurrency problem in Apache Sling. Therefore closing

PR #76, opened 2021-05-15
FELIX-6416 Fix url handling

PiotrKlimczak:
Not sure if is safe to have just url.getPath() as I am not Felix Framework expert.
But even if this PR is not suitable for merge, at least it has test coverage for the scenario encountered and rolls back the cause of it.
Would be good to back port that to 6.0.5 release which can be used by Karaf 4.3.X release as it is blocker for Narayana and possibly some other libs.
It is also blocking KARAF-7152.

But indeed it is rendering FELIX-6326 broken again AFAIK


karlpauls:
Closing in favor of #77 - many thanks for the test case!

PR #77, opened 2021-05-16
FELIX-6416: special case runtime ref in bundle urls

karlpauls:


PiotrKlimczak:
@karlpauls thanks for your PR.
It will definitely solve immediate problem.

Thanks

PR #79, opened 2021-06-10
FELIX-6430 Fixed runtime compatibility with JDK17 - prototype

dmatej:
- rather prototype of the fix, but works with current GlassFish's master branch + JDK17-ea+25-2252
- probably some steps can be removed
- logs stacktraces and messages, too much noise

I'm creating this PR rather for inspiration and I hope it can help to move forward more quickly ;)


karlpauls:
Thanks a lot @dmatej, I ended up using something similar in #80 - good idea! I think it should work now with the current trunk. Let me know if it still doesn't.

PR #81, opened 2021-07-21
Introduced thread-safety while updating field during deactivation phase

amitjoy:
The problem occurred when multiple injected services tried to be removed from the component's injection. It is highly likely because of the usage of `TreeMap` in such concurrent environment. Since `TreeMap` has never been written to be used in a concurrent environment, we should use `ConcurrentSkipListMap` instead as all the features of `TreeMap` have been introduced in `ConcurrentSkipListMap` and it is very much suitable for concurrent environment.

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/inject/field/FieldHandler.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/inject/field/FieldHandler.java
@@ -25,6 +25,7 @@ import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
 
@@ -42,7 +43,6 @@ import org.apache.felix.scr.impl.logger.ComponentLogger;
 import org.apache.felix.scr.impl.logger.InternalLogger.Level;
 import org.apache.felix.scr.impl.metadata.ReferenceMetadata;
 import org.osgi.framework.BundleContext;
-import org.osgi.service.log.LogLevel;
 
 /**
  * Handler for field references
@@ -186,15 +186,19 @@ public class FieldHandler
         	// unbind needs only be done, if reference is dynamic and optional
             if ( mType == METHOD_TYPE.UNBIND )
             {
-                if ( this.metadata.isOptional() && !this.metadata.isStatic() )
+                Map<RefPair<?, ?>, Object> boundValues = bp.getComponentContext().getBoundValues(metadata.getName());
+                synchronized (boundValues) 
                 {
-                    // we only reset if it was previously set with this value
-                    if ( bp.getComponentContext().getBoundValues(metadata.getName()).size() == 1 )
+                    if ( this.metadata.isOptional() && !this.metadata.isStatic() )
                     {
-                        this.setFieldValue(componentInstance, null);
+                        // we only reset if it was previously set with this value
+                        if ( boundValues.size() == 1 )
+                        {
+                            this.setFieldValue(componentInstance, null);
+                        }
                     }
+                    boundValues.remove(refPair);
                 }
-                bp.getComponentContext().getBoundValues(metadata.getName()).remove(refPair);
             }
             // updated needs only be done, if the value type is map or tuple
             // If it's a dynamic reference, the value can be updated
--- a/scr/src/main/java/org/apache/felix/scr/impl/manager/ComponentContextImpl.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/manager/ComponentContextImpl.java
@@ -19,7 +19,7 @@
 package org.apache.felix.scr.impl.manager;
 
 
-import java.util.Comparator;
+import java.util.Collections;
 import java.util.Dictionary;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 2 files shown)


amitjoy:
@tjwatson Could you kindly have a look at it?

tjwatson:
When the problem occurred what sort of errors/warnings or incorrect behavior did you observe?  Any stack trace?

amitjoy:
I have added the stacktrace to the JIRA bug report

scr/src/main/java/org/apache/felix/scr/impl/manager/ComponentContextImpl.java
```diff
@@ -328,7 +328,7 @@ public void dispose()
 
     private Map<RefPair<?, ?>, Object> createNewFieldHandlerMap()
     {
-        return new TreeMap<>(
+        return new ConcurrentSkipListMap<>(
```
tjwatson, line 331: While I think this will avoid the exception in TreeMap, I do not think it is a complete solution.  In `org.apache.felix.scr.impl.inject.field.FieldHandler.updateField(METHOD_TYPE, Object, BindParameters)` we have the following code:

```java
            if ( mType == METHOD_TYPE.UNBIND )
            {
                if ( this.metadata.isOptional() && !this.metadata.isStatic() )
                {
                    // we only reset if it was previously set with this value
                    if ( bp.getComponentContext().getBoundValues(metadata.getName()).size() == 1 )
                    {
                        this.setFieldValue(componentInstance, null);
                    }
                }
                bp.getComponentContext().getBoundValues(metadata.getName()).remove(refPair);
            }
```
`getBoundValues` call here will return this `ConcurrentSkipListMap`.  If multiple threads are entering this block then I think we have an issue with never calling `this.setFieldValue(componentInstance, null);` when two threads enter the `size() ==1` check and find the size to be 2.  Both threads will remove their own refPair from the Map but none will null out the field value when the last is removed.

cziegeler: Isn't the method that calls createNewFieldHandlerMap() synchronized?

cziegeler: Ignore the above comment :) But I think in general the code expects to not be called concurrently (if I remember correctly). So there should be some synchronization somewhere

amitjoy: @tjwatson According to the current implementation, `org.apache.felix.scr.impl.inject.field.FieldHandler.updateField(METHOD_TYPE, Object, BindParameters)` will be invoked by different implementations of `org.apache.felix.scr.impl.inject.field.FieldHandler.State.invoke(FieldHandler, METHOD_TYPE, Object, BindParameters)` and if we traverse back, none of the consumers of the aforementioned `invoke(..)` method is thread-safe (synchronized). I agree with you, we should introduce thread-safety in `org.apache.felix.scr.impl.inject.field.FieldHandler.updateField(METHOD_TYPE, Object, BindParameters)`.

amitjoy: Added a new commit to make the `org.apache.felix.scr.impl.inject.field.FieldHandler.updateField(METHOD_TYPE, Object, BindParameters)` synchronized.

scr/src/main/java/org/apache/felix/scr/impl/inject/field/FieldHandler.java
```diff
@@ -171,9 +171,9 @@ private boolean initField(final Object componentInstance,
         return objects;
     }
 
-    private MethodResult updateField(final METHOD_TYPE mType,
-                                     final Object componentInstance,
-                                     final BindParameters bp)
+    private synchronized MethodResult updateField(final METHOD_TYPE mType,
```
tjwatson, line 174: I think this synchronization is too broad.  In particular I am concerned about synchronization when making the calls to `org.apache.felix.scr.impl.inject.ValueUtils.getValue(String, ValueType, Class<?>, ScrComponentContext, RefPair<?, ?>)` which ultimately can set off a chain of getService calls that may trigger activating dependency components.

A more safe fix is to use a synchronized map and synchronize on the map around the bit of code that needs it:

```java
                Map<RefPair<?, ?>, Object> boundValues = bp.getComponentContext().getBoundValues(metadata.getName());
                synchronized (boundValues)
                {
                    if ( this.metadata.isOptional() && !this.metadata.isStatic() )
                    {
                        // we only reset if it was previously set with this value
                        if ( boundValues.size() == 1 )
                        {
                            this.setFieldValue(componentInstance, null);
                        }
                    }
                    boundValues.remove(refPair);
                }
```

Then I would just do `Collections.synchronizedMap(new TreeMap<>(...` when creating the map for bound values.

amitjoy: I saw it coming 😉 Yes, you are right. The `synchronized` block was a bit broader in this respect. We would be better off with synchronizing only the part that requires it. I will do the needful.

amitjoy: Applied your proposal

tjwatson (approved):
LGTM

amitjoy:
@tjwatson Is it possible to merge this as it can then be part of the next SCR release?

PR #83, opened 2021-07-30
FELIX-6441 Make it possible to run the Configuration Interpolation independently of Configuration Admin

bosschaert:
The new StandaloneInterpolator class provides the entry-point for using
the interpolator outside of a Configuration Admin environment.


cziegeler (approved):
Lgtm, thanks!

PR #84, opened 2021-08-03
FELIX-6441 - mention StandaloneInterpolator in README, and move test …

bdelacretaz:
…to verify that it works from a different package

@bosschaert I made some (nitpickers) changes here.


bdelacretaz:
Ok, sorry, I wasn't aware of that convention, the README changes are sufficient then.

bosschaert (approved):

PR #86, opened 2021-08-08
FELIX-6442 : NPE when generating ServiceReferenceDTOs

cziegeler:


karlpauls (approved):

framework/src/main/java/org/apache/felix/framework/DTOFactory.java
```diff
@@ -335,18 +335,27 @@ private static BundleStartLevelDTO createBundleStartLevelDTO(Bundle bundle)
...
+                dtos.add(dto);
+            }
         }
-        return dtos;
+        return dtos.toArray(new ServiceReferenceDTO[dtos.size()]);
     }
 
     private static ServiceReferenceDTO createServiceReferenceDTO(ServiceReference<?> svc)
     {
+        final Bundle bundle = svc.getBundle();
```
tjwatson, line 352: Note that the service properties also contain the bundle ID and must continue to return the bundle ID even after the service has been unregistered.

PR #87, opened 2021-08-09
FELIX-6442 : NPE when generating ServiceReferenceDTOs

cziegeler:


cziegeler:
@tjwatson I assume this is what you had in mind?

tjwatson:
> @tjwatson I assume this is what you had in mind?

Yes, that LGTM

PR #99, opened 2021-09-08
FELIX-6429 : IllegalArgumentException that should not happen

cziegeler:


cziegeler:
This one does not cover all cases

PR #114, opened 2021-11-05
Fix service-based and extender-based sample projects

andrewswan:
I was doing the Felix tutorials and noticed that the sample projects didn't work, so this PR fixes them.


cziegeler (approved):

cziegeler:
Thanks for providing this PR!

PR #121, opened 2021-11-26
FELIX-6473 - Make BundlesStartedCheck simpler

cschneider:
Use streams to make the check easier to understand.

### DIFF
--- a/healthcheck/generalchecks/src/main/java/org/apache/felix/hc/generalchecks/BundlesStartedCheck.java
+++ b/healthcheck/generalchecks/src/main/java/org/apache/felix/hc/generalchecks/BundlesStartedCheck.java
@@ -17,7 +17,10 @@
  */
 package org.apache.felix.hc.generalchecks;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
 
 import org.apache.felix.hc.annotation.HealthCheckService;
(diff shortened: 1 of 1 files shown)


healthcheck/generalchecks/src/main/java/org/apache/felix/hc/generalchecks/BundlesStartedCheck.java
```diff
@@ -79,78 +82,85 @@ protected void activate(BundleContext bundleContext, Config config) {
...
+                log.critical(msg);
+            } else {
+                log.warn(msg);
+            }
+        }
+        return isActive;
+    }
+
+    private String getBundleInfo(Bundle bundle) {
+        return String.format("bundle with id:%d, name:%s, state:%s)", bundle.getBundleId(),
```
ghenzler, line 159: Nitpick: the current format is just `Bundle {} {}: {}` (without labelling `id`/`name`/`state` on every line)
I think the values are self-explanatory in this context, I would keep it as is

healthcheck/generalchecks/src/main/java/org/apache/felix/hc/generalchecks/BundlesStartedCheck.java
```diff
@@ -79,78 +82,85 @@ protected void activate(BundleContext bundleContext, Config config) {
...
-            }
+    private boolean isNotFragment(Bundle bundle) {
+        boolean isFragment = StringUtils.isNotBlank(bundle.getHeaders().get(Constants.FRAGMENT_HOST));
+        if (isFragment) {
+            LOG.debug("Ignoring bundle fragment: {}", bundle.getSymbolicName());
         }
-
-        String includeMsg = !includesRegex.pattern().equals(".*") ? " for pattern " + includesRegex.pattern(): "";
-        String baseMsg = relevantBundlesCount + " bundles" + includeMsg;
-        String excludedMsg = countExcluded > 0 ? " (" + countExcluded + " excluded via pattern " + excludesRegex.get().pattern() + ")" : "";
```
ghenzler, line 139: Showing the exclude pattern in the result for the case it kicks in is quite useful IMHO - with this PR I only see `Bundles total: 578, expected to be active: 567, active: 566` and I don't immediately see the reason why 567 out of 578 are expected (someone might think this is due to the number 567 being configured when in reality it is an exclude regex)

healthcheck/generalchecks/src/main/java/org/apache/felix/hc/generalchecks/BundlesStartedCheck.java
```diff
@@ -79,78 +82,85 @@ protected void activate(BundleContext bundleContext, Config config) {
...
+        List<Bundle> bundles = Arrays.asList(bundleContext.getBundles());
+        AtomicLong checkedBundles = new AtomicLong();
+        long activeBundles = bundles.stream()
+                .filter(this::isIncluded)
+                .filter(this::isNotExcluded)
+                .filter(this::isNotFragment)
+                .filter(this::isNotLazyStarting)
+                .peek(bundle -> checkedBundles.incrementAndGet())
+                .filter(bundle -> isActive(log, bundle))
+                .count();
```
ghenzler, line 100: This stream approach is nice!

cziegeler:
Closing due to inactivity

PR #122, opened 2021-11-26
FELIX-6474 - Improve HealthCheckResultCacheTest

cschneider:
Use alternative constructor instead of spy.
Use local mock of  Servicereference.


ghenzler (approved):

ghenzler:
Looks cleaner thanks 👍

PR #129, opened 2022-02-07
Migrate ipojo to 11 jdk

cheesterx3:
There were solved several problems:

1. Using ASM5 - was increased to ASM7 for compiling in java 11+. For this change was removed `asm-all` dependency and used concrete dependencies
2. There was a problem on initializing final fields. iPOJO manipulator changed all class fields in constructor to initialize through setters invoking. And this was a problem because such possibility was fixed in java 9 ([Compilers accept modification of final fields outside initializer methods](https://bugs.openjdk.java.net/browse/JDK-8157181)). Besides there is no need to change final fields initialization, because after manipulator processing they are always accessed through getters
3. Was fixed static interface initialization

I had to change version to 1.13.0 line cause my own usage, i think it's good to change minor version if PR will be accepted

### DIFF
--- a/ipojo/manipulator/annotations/pom.xml
+++ b/ipojo/manipulator/annotations/pom.xml
@@ -25,7 +25,7 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>org.apache.felix.ipojo.annotations</artifactId>
-    <version>1.12.3</version>
+    <version>1.13.0-SNAPSHOT</version>
     <name>Apache Felix iPOJO Annotations</name>
 
     <description>
--- a/ipojo/manipulator/annotations/pom.xml
+++ b/ipojo/manipulator/annotations/pom.xml
@@ -25,7 +25,7 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>org.apache.felix.ipojo.annotations</artifactId>
-    <version>1.12.2-SNAPSHOT</version>
+    <version>1.12.3</version>
     <name>Apache Felix iPOJO Annotations</name>
 
     <description>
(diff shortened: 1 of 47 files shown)


gnodet:
@cheesterx3 I've tried running `mvn -Prelease install` but it fails.
Could you please investigate ?  This is necessary to perform the release.

cheesterx3:
> @cheesterx3 I've tried running `mvn -Prelease install` but it fails. Could you please investigate ? This is necessary to perform the release. I'm using
> 
> ```
> Apache Maven 3.9.5 (57804ffe001d7215b5e7bcb531cf83df38f93546)
> Maven home: /Users/gnodet/.sdkman/candidates/maven/3.9.5
> Java version: 1.8.0_362, vendor: BellSoft, runtime: /Users/gnodet/.sdkman/candidates/java/8.0.362-librca/jre
> Default locale: en_FR, platform encoding: UTF-8
> OS name: "mac os x", version: "14.1.1", arch: "aarch64", family: "mac"
> ```

I'll check it

cheesterx3:
@gnodet  I've made some fixes in `javadoc` plugin configuration. Now `mvn -Prelease install` works well

gnodet:
> @gnodet I've made some fixes in `javadoc` plugin configuration. Now `mvn -Prelease install` works well

Thx, that looks better, but not enough...
It still fails during the release process, you can check with `mvn release:prepare -DdryRun=true -DreleaseVersion=1.13.0 -DdevelopmentVersion=1.13.1-SNAPSHOT`
I don't think we need to fix all the javadoc, but we need to ignore those errors with the correct javadoc option...

gnodet:
@cheesterx3 thx, I pushed a few updates, could you double check it works on your side ?

cheesterx3:
> @cheesterx3 thx, I pushed a few updates, could you double check it works on your side ?

I've checked it. Looks correct, works as expected

PR #130, opened 2022-02-27
[FELIX-6494] upgrade bndlib from 5.1.1 to 6.2.0

hboutemy:
https://issues.apache.org/jira/browse/FELIX-6496
non-reproducible order of Exports/Imports https://github.com/bndtools/bnd/issues/5021 has been fixed

### DIFF
--- a/tools/maven-bundle-plugin/pom.xml
+++ b/tools/maven-bundle-plugin/pom.xml
@@ -68,7 +68,7 @@
     <plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-plugin-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


jbonofre:
@hboutemy thanks, I gonna take a look.

rotty3000:
FYI, the tradition for versioning the maven-bundle-plugin used to be to sync with the version of bndlib in order to more clearly notice that fact. Could we go back to that? It's only since the 5.1.2 that we stopped doing it (i.e. the last 2 releases).

@jbonofre if you don't plan to look at this soon I can.

jbonofre:
I'm on it ;)

PR #131, opened 2022-03-11
FELIX-6511: Cancel the SCR Component Registry's changeCountTimer on deactivation.

chrisr3:
Cancel the `changeCountTimer`. Also name its thread "SCR Component Registry" rather than leaving it as "Timer-0".

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/ComponentRegistry.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/ComponentRegistry.java
@@ -732,7 +732,7 @@ public class ComponentRegistry
             final Timer timer;
             synchronized ( this.changeCountTimerLock ) {
                 if ( this.changeCountTimer == null ) {
-                    this.changeCountTimer = new Timer();
+                    this.changeCountTimer = new Timer("SCR Component Registry", true);
                 }
                 timer = this.changeCountTimer;
             }
@@ -772,6 +772,13 @@ public class ComponentRegistry
                     "Service changecount Timer for {0} had a problem", e,
                     registration.getReference());
             }
+        }
+    }
+
+    public void shutdown() {
+        final Timer timer = changeCountTimer;
+        if (timer != null) {
+            timer.cancel();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


scr/src/main/java/org/apache/felix/scr/impl/ComponentRegistry.java
```diff
@@ -732,7 +732,7 @@ public void updateChangeCount()
             final Timer timer;
             synchronized ( this.changeCountTimerLock ) {
                 if ( this.changeCountTimer == null ) {
-                    this.changeCountTimer = new Timer();
+                    this.changeCountTimer = new Timer("SCR Component Registry");
```
tjwatson, line 735: Should we also make the `Timer` use a daemon thread here by passing true?

chrisr3: I don't know about "should", but we obviously _could_ do that too.

chrisr3: I've updated the `Timer` thread to be a daemon.

tjwatson: Thanks, my thought is that I cannot think of any reason we would want this thread to hold up the JVM from exiting.  The Framework itself is responsible to make sure there is at least one non-daemon thread active while the framework is running.  But once the framework has stopped I think the JVM should be free to exit, even in the off chance that this timer is still active.

scr/src/main/java/org/apache/felix/scr/impl/ComponentRegistry.java
```diff
@@ -774,4 +774,11 @@ public void run()
             }
         }
     }
+
+    public void shutdown() {
```
laeubi, line 778: Should this not be guarded with `synchronized ( this.changeCountTimerLock ) {` and probably set the field to null?

tjwatson: I don't think so.  `changeCountTimer` is volatile here and the object itself is thread safe.  Calling cancel multiple times should not be an issue.  Besides that once ComponentRegistry is shutdown references to it are gone from the activator and it should GC just fine.  Technically even the activator doesn't need to null out the registry there because the framework throws away the activator instance after the bundle is stopped.

Also, nulling out would incorrectly allow a new timer if the method `org.apache.felix.scr.impl.ComponentRegistry.updateChangeCount()` is somehow called again from another thread after the bundle has stopped.

PR #132, opened 2022-03-21
FELIX-6512 : Support for synchronous registration of persistence mana…

rotty3000:
…gers and configuration plugins

Signed-off-by: Raymond Augé <rotty3000@apache.org>


cziegeler:
@rotty3000  The patch looks good to me - I was wondering whether we need the configuration option? Or in other words, are there cases where sync registration does not work?
TBH I don't recall why I implemented this being async

rotty3000:
@cziegeler right. Could there be a risk in removing the existing behaviour? That was the question I was asking myself. However, if you wish I can certainly simplify this down to removing the async. We've been using it like this for literally years. On the other hand I would not consider our usage an all encompassing one.

cziegeler:
@rotty3000 I'll give it a test with the different Apache Sling launchers where I know this feature is used. I'll report back here

cziegeler:
@rotty3000 I did some tests with Apache Sling and did not run into problem with synchronous registration. So I think we can just switch and remove all code related to async registration.

rotty3000:
@cziegeler please see updated code

cziegeler (approved):
lgtm, thanks

Fool256 (approved):

Fool256 (approved):

PR #143, opened 2022-04-22
FELIX-6522 Empty name is not allowed in permissions.

dmatej:
- presumption: such capability can be removed as impliesDirect would return false too.


karlpauls:
Thanks for the patch!

PR #146, opened 2022-05-02
FELIX-6526|FELIX-6527 Handle absent incremental manifest data file

HannesWell:
This PR should fix [FELIX-6526](https://issues.apache.org/jira/browse/FELIX-6526) and [FELIX-6527](https://issues.apache.org/jira/browse/FELIX-6527).
The case that that the incremental manifest data file does not existing in regular, non-incremental builds was forgotten here and no unit-test for `ManifestPlugin` exists that could have detect that.


akurtakov:
@tjwatson is this for you? With dependabot and etc. uptake of new versions are faster than they used to be.

HannesWell:
It would be great if a maintainer of this plugin would have a look at this PR since the last three PRs address this issue.

jbonofre:
Thanks, I'm on it, heading to new maven-bundle-plugin release.

PR #148, opened 2022-05-04
FELIX-6528 Fixed missing cached data scenario

dmatej:


olamy:
duplicate for https://github.com/apache/felix-dev/pull/150
but maybe it's better to check if file exists rather than relying on exception?

dmatej:
> duplicate for #150 but maybe it's better to check if file exists rather than relying on exception?

Exactly, but the method I used was already there, so maybe the another usage should be resolved in the same way.

jbonofre:
Fixed by #146

PR #152, opened 2022-05-19
FELIX-6533 Fixed processing maven version from Felix.properties

dmatej:
- fixed by refactoring:
  - impl for processing system versions incl. classifier moved to Version
  - Version used for both OS and Felix version
  - to stay compatible with previous versions the original method was just
    marked as deprecated, can be removed later (public static)
- NativeLibraryClauseTest was incorrect for mac os (spaces)
- FrameworkVersionTest was changed to VersionTest


karlpauls:
Hi @dmatej, unfortunately, we can't move this into the osgi Version class as that is spec'ed api. So you'd have to move that back into the felix namespace. Furthermore, with all the changes, it is very hard to see what the actual fix you did is (and I'm not sure I found a description of what exactly is going wrong).  

Could you please create a PR that does the minimal change needed and then maybe a follow-up with refactorings?

dmatej:
The fix is this cleanup, result of it, fixed state, not some "fixed if condition". This is a minimized change, except some generics I left in one smaller class (NativeLibraryClause) and autoremoved whitespaces in Version (can be disabled in review). I can remove them too, but then it will take more of my time and we are just going back to swamps.

I don't understand why the specified API is included directly into the Felix framework module, but ok, I will move it somewhere else, despite it will reduce readability for developers.

dmatej:
Is it better now? I removed changes in generics and one for cycle, whitespaces, and moved the code out of the osgi packages to own class.

karlpauls:
Looks good, thanks

PR #156, opened 2022-06-02
FELIX-6523 BundlesStartedCheck status is not aligned with other configs in the General HC

deepakvarshney:
making bundleStartedCheck to return TEMPORARILY_UNAVAILABLE instead of WARN to make it consistent with other health checks


ankitaagar (approved):
LGTM

cziegeler (approved):

PR #168, opened 2022-07-18
FELIX-6551 | OSGi event blocklisting is not informative

akankshajain18:


eventadmin/impl/src/main/java/org/apache/felix/eventadmin/impl/handler/EventHandlerProxy.java
```diff
@@ -451,7 +451,7 @@ public void denyEventHandler()
     	{
 	        LogWrapper.getLogger().log(
 	                        LogWrapper.LOG_WARNING,
-	                        "Denying event handler from ServiceReference [" + this.reference + " | Bundle("
+	                        "Denying event handler from ServiceReference [" + this.handler.getClass().getCanonicalName() + " | Bundle("
```
cziegeler, line 454: I think we should still log the reference and log the classname in addition. Please note that there is a low chance that handler is already null (concurrent requests), so we need to guard against it.
We should also switch the log level from warning to error

akankshajain18: @cziegeler, incorporated review comments, please have a look

cziegeler (approved):

PR #170, opened 2022-08-05
[FELIX-6554] Fixed concurrent access to the internal data

amitjoy:
Since the bundles can appear/disappear dynamically, concurrent access is required for these data

Fixes [FELIX-6554](https://issues.apache.org/jira/browse/FELIX-6554)

### DIFF
--- a/configurator/src/main/java/org/apache/felix/configurator/impl/model/State.java
+++ b/configurator/src/main/java/org/apache/felix/configurator/impl/model/State.java
@@ -25,23 +25,22 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class State extends AbstractState implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
     /** Serialization version. */
-    private static final int VERSION = 1;
+    private static final int VERSION = 2;
 
     public static final String FILE_NAME = "state.ser";
 
-    private Map<Long, Long> bundlesLastModified = new HashMap<>();
-
-    private Map<Long, Long> bundlesConfigAdminBundleId = new HashMap<>();
+    private Map<Long, Long> bundlesLastModified = new ConcurrentHashMap<>();
+    private Map<Long, Long> bundlesConfigAdminBundleId = new ConcurrentHashMap<>();
 
     private volatile Set<String> initialHashes;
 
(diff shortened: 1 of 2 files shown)


amitjoy:
@cziegeler @tjwatson Could you kindly have a look?

configurator/src/main/java/org/apache/felix/configurator/impl/model/State.java
```diff
@@ -39,9 +39,8 @@ public class State extends AbstractState implements Serializable {
 
     public static final String FILE_NAME = "state.ser";
 
-    private Map<Long, Long> bundlesLastModified = new HashMap<>();
-
-    private Map<Long, Long> bundlesConfigAdminBundleId = new HashMap<>();
+    private Map<Long, Long> bundlesLastModified = new ConcurrentHashMap<>();
```
tjwatson, line 42: I don't get why the default implementation of serializable is not used here to read/write the maps so you can make these final fields.

tjwatson (commented):
This is the first I'm looking at this code.  But I am unsure why it needs special handling for serialization vs what the JVM would do for all the non-transient fields of the class.

configurator/src/main/java/org/apache/felix/configurator/impl/model/State.java
cziegeler: Its used for versioning of the serialized format, it makes it easier to make any change to this class in future versions. It might be a little bit overkill here

amitjoy: @cziegeler Could you kindly elaborate what I am missing from the PR?

tjwatson: The Map type changing is what triggered my comment about serialization.  The concern is that a previous data stream would be storing a HashMap instance and when you do the read on restoring it will construct a HashMap instance back.  So in that case you need to read into a Map local instance for the call to `Map<Long, Long>) in.readObject();` then check the instance of the Map.  If it is not a ConcurrentHashMap then you need to copy the content of the read map into a new ConcurrentHashMap instance.

But it seems there is other issues with thread safety for the `private volatile Set<String> initialHashes` set.  Is that always set to a ConcurrentHashSet or something similar to ensure it is thread safe?

cziegeler: Tom is right, you need to create the right type on deseralization.
But, it might also be that the synchronization is on the wrong level and instead of making the properties of this class thread safe take care of syncing on a higher level. I think, that was the intention, but I can't really remember the details

amitjoy: @tjwatson @cziegeler Thanks a lot for the clarifications 👍  It's my bad since I didn't notice it earlier. I will add the necessary changes and update the PR accordingly.

amitjoy: @tjwatson @cziegeler I have updated the PR. Could you kindly have another look?

cziegeler:
Thanks @amitjoy  - the PR by itself looks good now; but it is probably fixing the problem at the wrong place, see https://issues.apache.org/jira/browse/FELIX-6554

amitjoy:
The fix has already been introduced, therefore, closing it!

PR #172, opened 2022-09-14
[FELIX-6564] Fixed NPE while retrieving component descriptions from SCR

amitjoy:
Fixes https://issues.apache.org/jira/browse/FELIX-6564

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/runtime/ServiceComponentRuntimeImpl.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/runtime/ServiceComponentRuntimeImpl.java
@@ -117,6 +117,10 @@ public class ServiceComponentRuntimeImpl implements ServiceComponentRuntime
         try
         {
             ComponentHolder<?> holder = getHolderFromDescription( description);
+            // the holder can also be null if the associated component is deregistered
+            if (holder == null) {
+                return Collections.emptyList();
+            }
             // Get a fully filled out valid description DTO
             description = holderToDescription(holder);
             if ( description == null)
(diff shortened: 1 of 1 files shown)


amitjoy:
@tjwatson Could you kindly have a look?

amitjoy:
@tjwatson Do you have any update on this?

scr/src/main/java/org/apache/felix/scr/impl/runtime/ServiceComponentRuntimeImpl.java
```diff
@@ -117,6 +117,10 @@ public Collection<ComponentConfigurationDTO> getComponentConfigurationDTOs(Compo
         try
         {
             ComponentHolder<?> holder = getHolderFromDescription( description);
+            // the holder can also be null if the associated component is deregistered
```
tjwatson, line 120: If null is allowed to be returned here then I think we need null checks in all 4 locations it is called:

`disableComponent(ComponentDescriptionDTO)`
`enableComponent(ComponentDescriptionDTO)`
`getComponentConfigurationDTOs(ComponentDescriptionDTO)`
`isComponentEnabled(ComponentDescriptionDTO)`

Also, looking at `getHolderFromDescription(ComponentDescriptionDTO description)` I think it has an issue where it could get called on an uninstalled bundle in which case `Bundle b = context.getBundle(bundleId);` will result in a `null` value in `b` causing an NPE also.

tjwatson (approved):
LGTM, just want to confirm you ran the automated tests on this locally?

amitjoy:
I built it locally and all the tests ran successfully

amitjoy:
If you want, you can also cross-check it from your end. Just to be on the safe side.

tjwatson:
> If you want, you can also cross-check it from your end. Just to be on the safe side.

The fix looks safe enough, I just wanted to verify they ran on your end.

PR #174, opened 2022-09-22
Test for https://issues.apache.org/jira/browse/FELIX-6470

bosschaert:


bosschaert:
Closing this PR as consensus in the OSGi Spec Project is that the coordinator should not expose intermediate states - only the final state.

PR #177, opened 2022-10-13
FELIX-6570 - Components webconsole-plugin shows password in clear text

sagarmiglani:


webconsole-plugins/ds/src/main/java/org/apache/felix/webconsole/plugins/ds/internal/ConfigurationSupport.java
```diff
@@ -73,4 +75,16 @@ public boolean isConfigurable(final Bundle providingBundle, final String pid)
...
+
+    /**
+     * Returns a Collection of IDs of Password Attributes Definitions for given bundle and configuration PIDs
+     * @param bundle The Bundle providing the component
+     * @param configurationPids A non-null configuration pid
+     * @return <code>Collection<String></code>
+     */
+    public Collection<String> getPasswordAttributeDefinitionIds(final Bundle bundle, final String[] configurationPids) {
+        Object metaTypeService = this.metatypeTracker.getService();
+        return new MetatypeSupport().getPasswordAttributeDefinitionIds(metaTypeService, bundle, configurationPids);
```
cziegeler, line 87: I think its better to first check metaTypeService for null and only if not null call MetatypeSupport

cziegeler (approved):

webconsole-plugins/ds/src/main/java/org/apache/felix/webconsole/plugins/ds/internal/ConfigurationSupport.java
sagarmiglani: Thanks Carsten, added the new commit, please review the latest state.

PR #192, opened 2023-01-26
Updated to use HTTP optionally in HealthCheck

amitjoy:
fixes [FELIX-6565](https://issues.apache.org/jira/browse/FELIX-6565)

### DIFF
--- a/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
+++ b/healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
@@ -17,15 +17,23 @@
  */
 package org.apache.felix.hc.core.impl.servlet;
 
+import static org.osgi.service.http.whiteboard.HttpWhiteboardConstants.HTTP_WHITEBOARD_CONTEXT_SELECT;
+import static org.osgi.service.http.whiteboard.HttpWhiteboardConstants.HTTP_WHITEBOARD_SERVLET_PATTERN;
+
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Dictionary;
+import java.util.HashMap;
+import java.util.Hashtable;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.TreeMap;
 
+import javax.servlet.Servlet;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -38,12 +46,14 @@ import org.apache.felix.hc.api.execution.HealthCheckExecutor;
 import org.apache.felix.hc.api.execution.HealthCheckSelector;
 import org.apache.felix.hc.core.impl.executor.CombinedExecutionResult;
 import org.apache.felix.hc.core.impl.util.lang.StringUtils;
+import org.osgi.dto.DTO;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.ConfigurationPolicy;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
-import org.osgi.service.http.HttpService;
 import org.osgi.service.metatype.annotations.Designate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


amitjoy:
I just added a new commit to update to HTTP whiteboard. There was no runtime issue to use the old pattern of using `HttpService` to register the servlets, but dependency to `HttpService` introduces a mandatory requirement to `osgi.service` namespace. 

`osgi.service;filter:="(objectClass=org.osgi.service.http.HttpService)";effective:=active`

This requirement doesn't play any role in the runtime but it plays a signifact role of composing systems. If we use bndtools and use this bundle as an initial requirement to resolve the dependencies, this will fail as the `requirement` needs to be satisfied first. We also use similar mechanism to compose systems using OSGi Resolver which would not work if the comprises such mandatory dependency to `HttpService`. That's why, a cleaner approach would be to go for OSGi HTTP Whiteboard where dependency to `HttpService` is not required at all, whereas, it now optionally requires HTTP whiteboard.

amitjoy:
@ghenzler @cziegeler Could you kindly have a look at it?

amitjoy:
@jbonofre Thanks a lot for assigning yourself for the review 👍

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
```diff
@@ -162,7 +164,10 @@ void setDescription(String d) {
     ResultTxtVerboseSerializer verboseTxtSerializer;
 
     @Activate
-    protected final void activate(final HealthCheckExecutorServletConfiguration configuration) {
+    protected final void activate(final HealthCheckExecutorServletConfiguration configuration, final BundleContext bundleContext) {
+    	this.bundleContext = bundleContext;
+    	servletRegistrations = new ArrayList<>();
```
JochenHiller, line 169: Assign with `this.servletRegistrations`  to follow same code style?

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
```diff
@@ -215,29 +219,24 @@ protected final void activate(final HealthCheckExecutorServletConfiguration conf
...
-                this.httpService.unregister(servletPath);
-            } catch (Exception e) {
-                LOG.error("Could not unregister health check servlet: " + e, e);
-            }
-        }
-        this.servletPaths = null;
+        servletRegistrations.forEach(ServiceRegistration::unregister);
+    }
+    
+    private void registerServlet(Servlet servlet, String path) {
```
JochenHiller, line 234: Better to reverse argument ordering? Then it would match closer the `httpService.registerServlet()`.

E.g.

```
private void registerServlet(String path, Servlet servlet) {
```

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
```diff
@@ -215,29 +219,24 @@ protected final void activate(final HealthCheckExecutorServletConfiguration conf
...
-        for (final String servletPath : this.servletPaths) {
-            try {
-                LOG.info("Unregistering HC Servlet {} from path {}", getClass().getSimpleName(), servletPath);
-                this.httpService.unregister(servletPath);
-            } catch (Exception e) {
-                LOG.error("Could not unregister health check servlet: " + e, e);
-            }
-        }
-        this.servletPaths = null;
+        servletRegistrations.forEach(ServiceRegistration::unregister);
```
JochenHiller, line 231: Nice code, but the info that HC Servlet has been unregistered is missing compared to previous code. I would propose to have the logs like before so you see what has been registered and unregistered later.

Do you need to remove the entries from `servletRegistrations` or can we rely that no-one is using that variable after component de-activation?

JochenHiller (commented):
I like that change, as it changes the runtime dependency to `HttpService` to a build-time dependency only.

In our scenarios of using Apache Felix HC we do not use anymore `HttpService`, and we do not want to introduce runtime dependencies if not really necessary to save space in Embedded use cases.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
amitjoy: Yeah, makes sense. On it.

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
amitjoy: Yeah, agreed 👍

healthcheck/core/src/main/java/org/apache/felix/hc/core/impl/servlet/HealthCheckExecutorServlet.java
amitjoy: Yeah, since no one is using it, I didn't remove the entries. I will still clean it up for completeness.

JochenHiller:
Latest changes from @amitjoy LGTM

amitjoy:
@jbonofre if you have time, could you kindly have a look?

amitjoy:
@jbonofre Could you kindly have a look?

cziegeler (approved):

PR #193, opened 2023-02-06
FELIX-6591: Deallocate Conditions and Permissions when bundles unloaded.

chrisr3:
Fix memory leak when bundles are uninstalled from the Felix framework:
- Fix the "`equals()` / `hashCode()`" contract for the `Permissions.Entry` class when the entry it is wrapping is an array.
- Remove `Conditions` objects explicitly when their bundles are uninstalled, instead of using `WeakHashMap`s.

### DIFF
--- a/framework.security/src/main/java/org/apache/felix/framework/SecurityActivator.java
+++ b/framework.security/src/main/java/org/apache/felix/framework/SecurityActivator.java
@@ -20,6 +20,8 @@ package org.apache.felix.framework;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


karlpauls:
LGTM - thanks a lot for the patch @chrisr3

chrisr3:
> LGTM - thanks a lot for the patch @chrisr3

No problem. Do you have any idea when the next version of Felix Security will be released please?

karlpauls:
@chrisr3 - I'm planning to cut a release this week.

PR #194, opened 2023-02-13
Update Felix Log to latest org.osgi.service.log version (FELIX-6593)

glimmerveen:


glimmerveen:
I ran into an issue when updating a project's dependency to osgi.core to 8.0.0, that also used LogService. It appears that when org.osgi.service.log was moved from Compendium to Core it got a new version which has not been incorporated into  Felix Log yet.

This PR should address this by updating the org.osgi.service.log dependency to version 1.5. I also took the liberty to update the version to the next minor, as I noticed this was also done when the spec dependency was updated from 1.3 to 1.4 in the past.

glimmerveen:
@rotty3000 , I saw in the git history you incorporated a change related to OSGi Log version 1.5. Could you have a look at this PR, as I think this is needed to support at runtime code using OSGi Log, when it was compiled using `org.osgi:osgi.core:8.0.0`.

glimmerveen:
Thank you Raymond for looking into this PR and merging it!

Assuming there is no other work pending for this module, would it be possible to cut a 1.3.0 release of log?

PR #210, opened 2023-05-15
FELIX-6607 consider service ranking for conflicting web console plugins

kwin:


kwin:
@cziegeler @karlpauls  Can you have a look?

laeubi:
@kwin why not use a dummy service reference for internal plugins (or simply register the as regular services...)

kwin:
> @kwin why not use a dummy service reference for internal plugins (or simply register the as regular services...)

To be honest, because I see no need. I am fine with every external plugin overwrite internal ones with same labels. Making that dependent on service ranking doesn't seem useful.

cziegeler (approved):

PR #213, opened 2023-07-09
[Health Check] SFL4j 2.x compatibility

glimmerveen:
Currently Felix Health Check cannot be used in a runtime where SLF4j 2.x is deployed. I traced this to Health Check's API dependency on the `org.slf4j.helpers.MessageFormatter`. This PR replaces this dependency with a custom format implementation that should be equivalent to what was used from SLF4j API.

Some background: HC API currently builds against SLF4j 1.7.x, and uses both the api package itself and the MessageFormatter from the `org.slf4j.helpers` package. Within SLF4j 2.x, the main API package is exported in both 1.7.36 and 2.x, allowing this bundle to serve both SLF4j 1.7.x users and 2.x users. This "double version export" is not done for the helpers package, and thus the current HC API bundle cannot resolve when SLF4j 2.x is deployed.

### DIFF
--- a/healthcheck/api/bnd.bnd
+++ b/healthcheck/api/bnd.bnd
@@ -8,5 +8,6 @@ Bundle-License: Apache License, Version 2.0
 
 Bundle-Vendor: The Apache Software Foundation
 
-Export-Package: org.apache.felix.hc.api.*
+Export-Package: !org.apache.felix.hc.api.util,\
+    org.apache.felix.hc.api.*
 
--- a/healthcheck/api/src/main/java/org/apache/felix/hc/api/FormattingResultLog.java
+++ b/healthcheck/api/src/main/java/org/apache/felix/hc/api/FormattingResultLog.java
@@ -20,8 +20,8 @@ package org.apache.felix.hc.api;
 import java.text.NumberFormat;
 import java.util.Locale;
 
+import org.apache.felix.hc.api.util.FormatUtil;
 import org.osgi.annotation.versioning.ProviderType;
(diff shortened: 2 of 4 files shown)


kwin:
There is still ongoing discussions whether that is a mistake or a deliberate decision from SLF4J v2 in https://jira.qos.ch/browse/SLF4J-579.

glimmerveen:
> There is still ongoing discussions whether that is a mistake or a deliberate decision from SLF4J v2 in https://jira.qos.ch/browse/SLF4J-579.

I was not aware of this discussion. If slf4j-api is updated to export the `org.slf4j.helpers` package as 1.7.36 as well, then there is no direct need for this change. 😃 

I was doubting whether the types in the helpers package were intended to be used by users. I see that this question was also raised within the JIRA comments under the issue you referenced, but I don't see a clear answer.

glimmerveen:
> I was doubting whether the types in the helpers package were intended to be used by users. I see that this question was also raised within the JIRA comments under the issue you referenced, but I don't see a clear answer.

I did a quick search across projects hosted on GitHub, and there I already found almost 7000 matches for "import org.slf4j.helpers"; so regardless of the intent, it is used quite extensively across a lot of different projects.

I propose to await the outcome of the discussion within the SLF4j community. Due to the extend it is used (looking only at projects hosted on GitHub), I think it is reasonable that the `org.slf4j.helpers` package should get that 1.7.36 export as well.

glimmerveen:
This appears to have been resolved in SLF4j (tested with slf4j-api 2.0.12 and logback 1.5.0).

PR #214, opened 2023-07-17
FELIX-6616 : reduce timing window for binding dynamic greedy refs

tjwatson:


tjwatson:
After a prolong amount of testing this fix I think it is ready to merge.  As mentioned in the commit message the timing window still exists but this greatly reduces the chances of getting out of order binding calls.

PR #233, opened 2023-08-30
FELIX-6515: Add global write lock to write back cfg files

FelixMarxIBM:
* Writing a file is a 2 step process where the file is created empty and then filled
* There is a race-condition, where the DirectoryWatcher might pick up the creation of the
* empty file. This empty file will be written back to disk on a follow-up CM_UPDATED event.
* Using the global write lock will block the DirectoryWatcher to wait for the write to finish
* and therefore never to read an empty file.


FelixMarxIBM:
I'll try to create a test for this race-condition but doing that might be hard to accomplish.

PR #236, opened 2023-09-17
AutoProcessor - Load and start bundles in parallel

pskowronek:
A PoC of loading and starting bundles in parallel in AutoProcessor.

The was tried in a project https://github.com/mucommander/mucommander that is using AutoProcessor to load bundles. For muCommander this gave 2-3 times faster bundle loading etc (from 2-3s to ~400-800ms on my macbook 2012) - see here https://github.com/mucommander/mucommander/pull/1044/commits/ab376b984257e79a563a0fa49cc128af9c501e9f (PR in muCommander: https://github.com/mucommacommits/ab376b984257e79a563a0fa49cc128af9c501e9f).

Can you please review and asses whether the concept could be incorporated into Felix, whether it is safe etc.

### DIFF
--- a/main/src/main/java/org/apache/felix/main/AutoProcessor.java
+++ b/main/src/main/java/org/apache/felix/main/AutoProcessor.java
@@ -20,6 +20,12 @@ package org.apache.felix.main;
 
 import java.io.File;
 import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
 import org.osgi.framework.*;
 import org.osgi.service.startlevel.*;
 
@@ -89,9 +95,17 @@ public class AutoProcessor
     **/
     public static void process(Map configMap, BundleContext context)
     {
-        configMap = (configMap == null) ? new HashMap() : configMap;
-        processAutoDeploy(configMap, context);
-        processAutoProperties(configMap, context);
+        ExecutorService executor = null;
+        try {
+            executor = Executors.newFixedThreadPool(8);
+            Map configMapLocal = (configMap == null) ? new HashMap() : configMap;
+            processAutoDeploy(executor, configMapLocal, context);
+            processAutoProperties(executor, configMapLocal, context);
+        } finally {
+            if (executor != null) {
+                executor.shutdown();
+            }
+        }
     }
 
     /**
@@ -100,7 +114,8 @@ public class AutoProcessor
      * specified deploy actions.
      * </p>
      */
-    private static void processAutoDeploy(Map configMap, BundleContext context)
+    private static void processAutoDeploy(ExecutorService executor, Map configMap,
+                                          BundleContext context)
     {
         // Determine if auto deploy actions to perform.
         String action = (String) configMap.get(AUTO_DEPLOY_ACTION_PROPERTY);
@@ -156,28 +171,28 @@ public class AutoProcessor
             // Look in the specified bundle directory to create a list
             // of all JAR files to install.
             File[] files = new File(autoDir).listFiles();
-            List jarList = new ArrayList();
+            List<File> jarList = new ArrayList<>();
             if (files != null)
             {
                 Arrays.sort(files);
-                for (int i = 0; i < files.length; i++)
+                for (File file : files)
                 {
-                    if (files[i].getName().endsWith(".jar"))
+                    if (file.getName().endsWith(".jar"))
                     {
-                        jarList.add(files[i]);
+                        jarList.add(file);
                     }
                 }
             }
 
             // Install bundle JAR files and remember the bundle objects.
-            final List startBundleList = new ArrayList();
-            for (int i = 0; i < jarList.size(); i++)
+            final List<Future<Bundle>> startBundleList = new ArrayList<>();
+            for (File jarFile : jarList)
             {
                 // Look up the bundle by location, removing it from
                 // the map of installed bundles so the remaining bundles
                 // indicate which bundles may need to be uninstalled.
-                Bundle b = (Bundle) installedBundleMap.remove(
-                    ((File) jarList.get(i)).toURI().toString());
+                Bundle b = (Bundle) installedBundleMap.remove(jarFile.toURI().toString());
+                Future<Bundle> futureBundle = CompletableFuture.completedFuture(b);
 
                 try
                 {
@@ -185,8 +200,18 @@ public class AutoProcessor
                     // if the 'install' action is present.
                     if ((b == null) && actionList.contains(AUTO_DEPLOY_INSTALL_VALUE))
                     {
-                        b = context.installBundle(
-                            ((File) jarList.get(i)).toURI().toString());
+                        String jarPath = jarFile.toURI().toString();
+                        b = context.installBundle(jarFile.toURI().toString());
+                        futureBundle = executor.submit(() -> {
+                            try {
+                                return context.installBundle(jarPath);
+                            } catch (BundleException ex) {
+                                System.err.println("Auto-deploy install: "
+                                        + ex + ((ex.getCause() != null) ? " - "
+                                        + ex.getCause() : ""));
+                                return null;
+                            }
+                        });
                     }
                     // If the bundle is already installed, then update it
                     // if the 'update' action is present.
@@ -194,15 +219,9 @@ public class AutoProcessor
                     {
                         b.update();
                     }
-
                     // If we have found and/or successfully installed a bundle,
                     // then add it to the list of bundles to potentially start
-                    // and also set its start level accordingly.
-                    if ((b != null) && !isFragment(b))
-                    {
-                        startBundleList.add(b);
-                        sl.setBundleStartLevel(b, startLevel);
-                    }
+                    startBundleList.add(futureBundle);
                 }
                 catch (BundleException ex)
                 {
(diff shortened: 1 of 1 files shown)


paulrutter:
@pskowronek interesting concept.
We've experimented with parallel startup of bundles as well, but got stuck with non-reproducible edge cases in which OSGi services would not start. Some kind of deadlock scenario.

Did you encounter this while creating this PR?

pskowronek:
> @pskowronek interesting concept. We've experimented with parallel startup of bundles as well, but got stuck with non-reproducible edge cases in which OSGi services would not start. Some kind of deadlock scenario.
> 
> Did you encounter this while creating this PR?

No, I didn't encounter this. We're using this patch in https://github.com/mucommander/mucommander project since around 6 months, and so far no deadlocks were reported.

paulrutter:
> > @pskowronek interesting concept. We've experimented with parallel startup of bundles as well, but got stuck with non-reproducible edge cases in which OSGi services would not start. Some kind of deadlock scenario.
> > Did you encounter this while creating this PR?
> 
> No, I didn't encounter this. We're using this patch in https://github.com/mucommander/mucommander project since around 6 months, and so far no deadlocks were reported.

Ok, that's good. I might try it as well in our product and see what happens. I think we had the issue 1 out of 10 startups, and have a script that i can test it with.
I will report back with my findings if i get to it.
It would be an interesting change, as it reduces startup time of the application.

paulrutter:
Without having reviewed the full PR, I think it would make sense to at least make this behavior optional and thus configurable, to not break any existing applications.
If users would opt-in to parallel startup, it's at least a conscious decision.

But let me first try it out in my application to see if i run into any issues.

pskowronek:
> Without having reviewed the full PR, I think it would make sense to at least make this behavior optional and thus configurable, to not break any existing applications. If users would opt-in to parallel startup, it's at least a conscious decision.

Yes, it should be configurable.

> But let me first try it out in my application to see if i run into any issues.

 Superb! Thanks.

stbischof:
Could you please run this against the OSGi TCKs

paulrutter:
Good point @stbischof.

stbischof:
This all happens in main Project. The  command line launcher.  So this may make it easier. But the startlevel chapters should be respected. Did you read them carefully?

paulrutter:
@pskowronek i checked how we ran parallel startup earlier, and it was based on code shared in https://felix.apache.org/documentation/subprojects/apache-felix-dependency-manager/reference/thread-model.html.

How does this PR work differently than via the dependency manager as outlined above?

paulrutter:
@pskowronek i tried it out and compared startup times against the "regular" build.
In the big scheme of things, i doesn't seem to make a big difference (our application is probably a lot larger than yours, startup times of several minutes are not unusual). Loading/starting the bundles in parallel didn't seem to impact our startup times much, but again, for smaller applications it could make a (positive) difference though.

I also running our script to detect startup issues, will let you know the outcome.

pskowronek:
> @pskowronek i tried it out and compared startup times against the "regular" build. In the big scheme of things, i doesn't seem to make a big difference (our application is probably a lot larger than yours, startup times of several minutes are not unusual). Loading/starting the bundles in parallel didn't seem to impact our startup times much, but again, for smaller applications it could make a (positive) difference though.
> 
> I'm also running our script to detect startup issues, will let you know the outcome.

Can you tell me how many bundles the test app has, and what was the processor you were testing on?

paulrutter:
I don't have the exact count at hand, but in the order of 30-40 bundles i think.
I used your branch to test it with, see https://github.com/apache/felix-dev/pull/236.

pskowronek:
> I don't have the exact count at hand, but in the order of 30-40 bundles i think. I used your branch to test it with, see [blueconic@7ac245c](https://github.com/blueconic/felix-dev/commit/7ac245c3dfaebe871bc3abcdb1289c7f32dff806).

ok, what type of processor the test was run on - the number of (HT) cores might matter. In muCommander we have around 180 bundles , and since this is desktop app, it cpu have 4 (8 HT) cores at minimum.

pderop:
@paulrutter ,

> I assume this PR just starts the OSGi bundles in parallel, where the above starts the services in these bundles in parallel, correct?

Hi Paul,

yes, you are correct, when DM is used in concurrent mode, then services are all started and registered concurrently, so concurrency is achieved at the service level.

now, about the exception you reported:

> Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.NullPointerException [in thread "pool-3-thread-1"]
>         at org.apache.felix.dm.impl.index.ServiceRegistryCacheManager.init(ServiceRegistryCacheManager.java:140)
>         at org.apache.felix.dm.impl.index.ServiceRegistryCacheManager.<clinit>(ServiceRegistryCacheManager.java:63)
>         at org.apache.felix.dm.DependencyManager.createContext(DependencyManager.java:314)
>         at org.apache.felix.dm.DependencyManager.<init>(DependencyManager.java:88)
>         at org.apache.felix.dm.DependencyActivatorBase.start(DependencyActivatorBase.java:77)
>         at org.apache.felix.framework.util.SecureAction.startActivator(SecureAction.java:849)
>         at org.apache.felix.framework.Felix.activateBundle(Felix.java:2429)
>         at org.apache.felix.framework.Felix.startBundle(Felix.java:2335)
>         at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:1006)
>         at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:992)
>         at org.apache.felix.main.AutoProcessor.lambda$processAutoProperties$2(AutoProcessor.java:376)
>         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
>         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
>         ... 1 more

after many attempts, I confirm that I also reproduced the problem you reported.
I'm not using DM in parallel mode, and I have recompiled the framework from this PR branch.
I'm using 117 bundles, on mac M1 with 8 cpus. The issue is hard to reproduce.

So, from the stacktrace, we have a NPE in this [location](https://github.com/apache/felix-dev/blob/master/dependencymanager/org.apache.felix.dependencymanager/src/org/apache/felix/dm/impl/index/ServiceRegistryCacheManager.java#L140), meaning that in the previous line ([here](https://github.com/apache/felix-dev/blob/master/dependencymanager/org.apache.felix.dependencymanager/src/org/apache/felix/dm/impl/index/ServiceRegistryCacheManager.java#L139)), then `bundle.getBundleContext();` has returned null.

I cannot comment on this for the moment.

stbischof:
@pskowronek 

It it correct that all your bubdles use onlY bundleactivators and no declarative services?

pskowronek:
> @pskowronek
> 
> It it correct that all your bubdles use onlY bundleactivators and no declarative services?

Yes

paulrutter:
> @paulrutter ,
> 
> > I assume this PR just starts the OSGi bundles in parallel, where the above starts the services in these bundles in parallel, correct?
> 
> Hi Paul,
> 
> yes, you are correct, when DM is used in concurrent mode, then services are all started and registered concurrently, so concurrency is achieved at the service level.
> 
> now, about the exception you reported:
> 
> > Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.NullPointerException [in thread "pool-3-thread-1"]
> > at org.apache.felix.dm.impl.index.ServiceRegistryCacheManager.init(ServiceRegistryCacheManager.java:140)
> > at org.apache.felix.dm.impl.index.ServiceRegistryCacheManager.(ServiceRegistryCacheManager.java:63)
> > at org.apache.felix.dm.DependencyManager.createContext(DependencyManager.java:314)
> > at org.apache.felix.dm.DependencyManager.(DependencyManager.java:88)
> > at org.apache.felix.dm.DependencyActivatorBase.start(DependencyActivatorBase.java:77)
> > at org.apache.felix.framework.util.SecureAction.startActivator(SecureAction.java:849)
> > at org.apache.felix.framework.Felix.activateBundle(Felix.java:2429)
> > at org.apache.felix.framework.Felix.startBundle(Felix.java:2335)
> > at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:1006)
> > at org.apache.felix.framework.BundleImpl.start(BundleImpl.java:992)
> > at org.apache.felix.main.AutoProcessor.lambda$processAutoProperties$2(AutoProcessor.java:376)
> > at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
> > at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
> > ... 1 more
> 
> after many attempts, I confirm that I also reproduced the problem. I'm not using DM in parallel mode, and I have recompiled the framework from this PR branch. I'm using 117 bundles, on mac M1 with 8 cpus. The issue is hard to reproduce.
> 
> So, from the stacktrace, we have a NPE in this [location](https://github.com/apache/felix-dev/blob/master/dependencymanager/org.apache.felix.dependencymanager/src/org/apache/felix/dm/impl/index/ServiceRegistryCacheManager.java#L140), meaning that in the previous line ([here](https://github.com/apache/felix-dev/blob/master/dependencymanager/org.apache.felix.dependencymanager/src/org/apache/felix/dm/impl/index/ServiceRegistryCacheManager.java#L139)), `bundle.getBundleContext();` has returned null.
> 
> I cannot comment on this for the moment.

I don't know what causes this, but maybe adding a while loop on the bundleContext being non null with a thread.sleep(100) would resolve this intermittent timing issue?
Then again, there might be more places where there could occur similar timing issues.
What do you think would be the best approach? The reason I'm interested in this is that we experienced the exact same issue sporadically when experimenting with parallel startup via the DM (hence the script i had laying around).

PR #241, opened 2023-10-25
FELIX-6664 - Replacing String representation with customized Buffered…

DominikSuess:
…Reader (newline behavior slightly changing (newlines inside comments eliminated)

### DIFF
--- a/cm.json/src/test/java/org/apache/felix/cm/json/io/impl/JsonSupportTest.java
+++ b/cm.json/src/test/java/org/apache/felix/cm/json/io/impl/JsonSupportTest.java
@@ -190,8 +190,6 @@ public class JsonSupportTest {
                 + "  \"a\" : 1,\n"
                 + "  \n"
                 + "  \n"
-                + "\n"
-                + "\n"
                 + "  \"b\" : 2\n"
                 + "}\n", parse(input));
     }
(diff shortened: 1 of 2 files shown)


rombert:
@DominikSuess - I think Carsten should review this. In the meantime, I see that the indentation is a bit off, perhaps you can look into this?

DominikSuess:
> @DominikSuess - I think Carsten should review this. In the meantime, I see that the indentation is a bit off, perhaps you can look into this?

Ah looks like I  accidentially moved the whole block left - fixed now.

cm.json/src/test/java/org/apache/felix/cm/json/io/impl/JsonSupportTest.java
```diff
@@ -190,8 +190,6 @@ public void testSeveralComments() throws IOException {
                 + "  \"a\" : 1,\n"
                 + "  \n"
                 + "  \n"
-                + "\n"
```
cziegeler, line 193: This seems to be a change in behaviour as line numbers will change after comments are removed

DominikSuess: fixed to retain the linecount

cziegeler (approved):
lgtm

PR #250, opened 2023-11-28
FELIX-6672 - rewrite of detection logic to save previous char, resolv…

DominikSuess:
…ing patterns (esp) start/stop tokens (//, /*, */) spreading across two reads.

### DIFF
--- a/cm.json/src/main/java/org/apache/felix/cm/json/io/impl/JsonSupport.java
+++ b/cm.json/src/main/java/org/apache/felix/cm/json/io/impl/JsonSupport.java
@@ -244,6 +244,8 @@ public class JsonSupport {
         private boolean insideComment = false;
         private boolean insideLineComment = false;
         private boolean insideString = false;
+        private boolean isSkippedSlash = false;
+        private char oldChar = 0; // priming with 0 as it is not part of comment or string escaping chars
             
         public CommentRemovingReader(Reader reader) {
             super(new BufferedReader(reader));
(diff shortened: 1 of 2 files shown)


rombert (approved):
Did not review in depth, but since we have additional tests we should be good.

cm.json/src/test/java/org/apache/felix/cm/json/io/impl/JsonSupportTest.java
```diff
@@ -307,4 +307,19 @@ public void testEscapedQuotes() throws IOException {
                 + "}\n", parse(input));
     }
 
+    @Test
+    public void testSlashAtEndOfRead() throws Exception {
+        String input = "This is a test string // Next string \n"
+                + "next line" ;
+        Reader reader = JsonSupport.createCommentRemovingReader(new StringReader(input));
+
+        char[] buffer = new char[22];
```
jsedding, line 316: Nit: Instead of the magic number 22, how about using `input.indexOf('/')`? Or something equivalent, if I misunderstood the intent. IMHO that would make it clearer what you're testing.

cm.json/src/main/java/org/apache/felix/cm/json/io/impl/JsonSupport.java
```diff
@@ -268,84 +270,109 @@ public int read(char[] cbuf, int off, int len) throws IOException {
...
+                filteredContent.append(currentLine.toString());
 
-                    char[] filteredChars = filteredContent.toString().toCharArray();
-                    int filteredLen = Math.min(filteredChars.length, len);
-                    System.arraycopy(filteredChars, 0, cbuf, off, filteredLen);
-                    return filteredLen;
-                }
-                return charsRead;
+                char[] filteredChars = filteredContent.toString().toCharArray();
+                int filteredLen = Math.min(filteredChars.length, len);
```
jsedding, line 359: Wouldn't it be a bug if `filteredChars.length > len`? If so, maybe throwing an exception would be better than smoothing over it.

PR #290, opened 2024-01-16
Adds Requirement for Configuration Admin

juergen-albert:
The Configurator has no requirement for the Configuration Admin, because without it it, the bundle is useless.

### DIFF
--- a/configurator/pom.xml
+++ b/configurator/pom.xml
@@ -74,6 +74,9 @@
                         <Provide-Capability>
                             osgi.extender;osgi.extender="osgi.configurator";version:Version="1.0"
                         </Provide-Capability>
+                        <Require-Capability>
+                            osgi.implementation;filter:='(&(osgi.implementation=osgi.cm)(version>=1.6.0)(!(version>=2.0.0)))'
+                        </Require-Capability>    
                     </instructions>
                 </configuration>
             </plugin>


juergen-albert:
@cziegeler I'm not sure if I set the CM Version to high in the requirement.

cziegeler:
The version is fine, it matches the imports and that version is required for the configurator to work.
Now, I'm not against adding this, but isn't usually the import package enough? I understand that in theory just an api bundle is sufficient to satisfy the import package, but then the API is useless.

juergen-albert:
It was in the Past, as the felix cm shipped the api packages as well. Today it does not, which leads to resolver to include the API bundle, but no implementation, which is quite annoying.

cziegeler:
Apache Felix ConfigAdmin still includes the api packages.
If it helps, we can include the require capability; however I think the api bundle is the one that should be fixed - as the API really requires an implementation. An API bundle alone is useless. That's easier to change then each and every consumer bundle of the API.

juergen-albert:
True that. I have opened an issue for that: https://github.com/osgi/osgi/issues/665

The Configurator should have the Requirement regardless, as it definitely requires an implementation. I'd be happy if you would merge this.

juergen-albert:
Thanks heaps!

configurator/pom.xml
```diff
@@ -74,6 +74,9 @@
                         <Provide-Capability>
                             osgi.extender;osgi.extender="osgi.configurator";version:Version="1.0"
                         </Provide-Capability>
+                        <Require-Capability>
+                            osgi.implementation;filter:='(&(osgi.implementation=osgi.cm)(version>=1.6.0)(!(version>=2.0.0)))'
```
laeubi, line 78: You need to encode entities in XML!

--> `osgi.implementation;filter:='(&amp;(osgi.implementation=osgi.cm)(version>=1.6.0)(!(version>=2.0.0)))'`

laeubi:
>  as the API really requires an implementation. An API bundle alone is useless. 

I think this statement is wrong. An API *never* requires an implementation it only defines a _contract_. Also claiming that an API alone is useless is not true, because services can be _optional_ in OSGi.

In fact only the consumer of the API can know if an _implementation_ is **mandatory** or **optional** for its function, for example a Web UI for configuring components using Http Whiteboard does not **require** either _Http Whiteboard_ Implementation nor _Config Admin Service_ it will just do nothing then so it could make sense to make it **mandatory** so the bundle will supply function and the management agent of the system can make the user aware.

On the other hand there can be components that just would add additional functions (e.g. SCR) if a ConfigAdmin is there, still they will need to implement the API and use it without not working at all without it, they will still provide their main purpose.

cziegeler:
I agree that there is a case by case base; I was purely talking about the configuration admin api which doesn't do anything if you don't have the service.
After 20 years of configuration admin with thousands of bundles using it, requiring each and every using bundle to have this require capability in their bundle does not seem right for me.
But regardless this is a discussion that should happen in the respective projects.

cziegeler:
fair enough. thanks

timothyjward:
Shouldn't this PR have added a service requirement for configuration admin rather than an implementation requirement for the whiteboard? The configurator requires one or more `ConfigurationAdmin` services to run, but as far as I am aware it doesn't use the whiteboard at all. Also should the requirement be effective at resolve time? It feels more like active time effectiveness to me (which would tie in with the service requirement).

configurator/pom.xml
timothyjward: I think that would be:
`osgi.implementation;filter:='(&amp;(osgi.implementation=osgi.cm)(version&gt;=1.6.0)(!(version&gt;=2.0.0)))'`

Perhaps better to use an annotation in the source for safety?

juergen-albert: This project still uses the bundle plugin. I wasn't sure how recent the BND implementation behind it is. Thus I didn't use the Annotation. Would be preferable though.

juergen-albert:
If no effective is declared, the default is resolve AFAIK. Thus this should do the job (I also copied the requirement produced by the annotation).

Not sure what you mean by the whiteboard here. To be correct the service requirement would make sense as well. From my perspective both would be correct.

rotty3000:
> This project still uses the bundle plugin.

The latest maven-bundle-plugin uses bndlib 6.3.1 which supports all the annotations to my knowledge (I don't believe there are new annotations since that time frame).

PR #296, opened 2024-02-22
Fix ModuleCapability/ModuleRequirement hashcode/equals according to API

laeubi:
The API of Capability/Requirement currently defines the contract to be: This Capability/Requirement is equal to another Capability/Requirement if they have the same namespace, directives and attributes and are declared by the same resource.

But Felix currently implements hashCode/equals in terms of object identity.

This is the counterpart for
- https://github.com/eclipse-equinox/equinox/pull/537

### DIFF
--- a/framework/src/main/java/org/apache/felix/framework/wiring/BundleCapabilityImpl.java
+++ b/framework/src/main/java/org/apache/felix/framework/wiring/BundleCapabilityImpl.java
@@ -24,16 +24,17 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


stbischof:
@tjwatson do we have some ci here? 

do we may need an github workflow

tjwatson:
> @tjwatson do we have some ci here?

Not sure what you are asking here.  Probably a question for @karlpauls 

> @laeubi do we have an osgi-tck test that coverts this behavier?

Need caution here as I remember Richard Hall and I designing the resource and resolver APIs such that the framework can use identity equality for capabilities and bundle revisions.

PR #298, opened 2024-02-26
Add jetty websocket support to Jetty12 

paulrutter:
This change makes it possible to use [Jetty WebSocket](https://eclipse.dev/jetty/documentation/jetty-12/programming-guide/index.html#pg-server-websocket-jetty) (not JakartaEE 10) functionality via the Felix HTTP Jetty 12 bundle.

There are still some strange things that need to be done when writing a WebSocket servlet to get it all to work, but that's for a separate change request/bug report. I requested access to the Felix Jira so i can add it to the jira project.

More info about the required changes needed for setting up an actual websocket servlet. I combined the solutions mentioned in the links below (change classloader, delayed init and Jetty container websocket initialization code), otherwise the Websocket servlet fails to be initialized. 
* https://www.eclipse.org/forums/index.php/t/1110140/
* https://issues.apache.org/jira/browse/FELIX-5310

`jetty-websocket-jetty-client` is technically not required, but handy to have available when writing client code (server to server websocket proxy for example). 

@cziegeler for visibility.

### DIFF
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -41,7 +41,7 @@
 
     <properties>
         <felix.java.version>17</felix.java.version>
-        <jetty.version>12.0.6</jetty.version>
+        <jetty.version>12.0.7</jetty.version>
         <baseline.skip>true</baseline.skip>
     </properties>
 
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -433,11 +433,6 @@
            <version>${jetty.version}</version>
         </dependency>
         <dependency>
-          <groupId>org.eclipse.jetty.websocket</groupId>
-          <artifactId>jetty-websocket-jetty-client</artifactId>
-          <version>${jetty.version}</version>
-        </dependency>
-        <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.service.servlet</artifactId>
             <version>2.0.0</version>
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -37,7 +37,6 @@
         <connection>scm:git:https://github.com/apache/felix-dev.git</connection>
         <developerConnection>scm:git:https://github.com/apache/felix-dev.git</developerConnection>
         <url>https://gitbox.apache.org/repos/asf?p=felix-dev.git</url>
-        <tag>org.apache.felix.http.jetty12-1.0.3</tag>
     </scm>
 
     <properties>
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -41,7 +41,7 @@
 
     <properties>
         <felix.java.version>17</felix.java.version>
-        <jetty.version>12.0.7</jetty.version>
+        <jetty.version>12.0.8</jetty.version>
         <baseline.skip>true</baseline.skip>
     </properties>
 
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -30,7 +30,7 @@
     <description>This is an implementation of the R8.1 OSGi Servlet Service, the R7 OSGi Http Service and the R7 OSGi Http Whiteboard Specification</description>
 
     <artifactId>org.apache.felix.http.jetty12</artifactId>
-    <version>1.0.3_BC</version>
+    <version>1.0.3-SNAPSHOT</version>
     <packaging>bundle</packaging>
 
     <scm>
--- a/http/jetty12/pom.xml
+++ b/http/jetty12/pom.xml
@@ -30,19 +30,19 @@
     <description>This is an implementation of the R8.1 OSGi Servlet Service, the R7 OSGi Http Service and the R7 OSGi Http Whiteboard Specification</description>
 
     <artifactId>org.apache.felix.http.jetty12</artifactId>
-    <version>1.0.3-SNAPSHOT</version>
+    <version>1.0.3_BC</version>
     <packaging>bundle</packaging>
 
     <scm>
         <connection>scm:git:https://github.com/apache/felix-dev.git</connection>
         <developerConnection>scm:git:https://github.com/apache/felix-dev.git</developerConnection>
         <url>https://gitbox.apache.org/repos/asf?p=felix-dev.git</url>
-        <tag>org.apache.felix.http.jetty12-1.0.2</tag>
+        <tag>org.apache.felix.http.jetty12-1.0.3</tag>
     </scm>
 
     <properties>
         <felix.java.version>17</felix.java.version>
-        <jetty.version>12.0.5</jetty.version>
+        <jetty.version>12.0.6</jetty.version>
         <baseline.skip>true</baseline.skip>
     </properties>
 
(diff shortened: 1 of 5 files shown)


http/jetty12/pom.xml
```diff
@@ -181,6 +181,9 @@
                             org.apache.commons.*
                         </Conditional-Package>
                         <Import-Package>
+                            jakarta.annotation.*;resolution:=optional,
```
paulrutter, line 184: These additions were needed, otherwise the bundle would not start because of missing dependencies. Not sure why though, as the Jetty websocket bundles should be Jakarta agnostic.

http/jetty12/pom.xml
```diff
@@ -181,6 +181,9 @@
...
+                            jakarta.transaction.*;resolution:=optional,
+                            org.objectweb.asm.*;resolution:=optional,
```
paulrutter, line 186: Same for this one, without this optional resoluton, the bundle won't start.

cziegeler:
Thanks @paulrutter . Did you try installing the required bundles in addition to the Apache Felix jetty12 bundle?
I'm not totally against embedding this support, but would prefer having it optional.

paulrutter:
@cziegeler yes, i tried that. But it's quite hard to determine which bundles you need, with all the transitive (and partly overlapping) dependencies.
See for example https://repo1.maven.org/maven2/org/eclipse/jetty/websocket/jetty-websocket-jetty-server/12.0.6/jetty-websocket-jetty-server-12.0.6.pom.

I could not get all those separate bundles to start, but i might not have the tried with the right combination of bundles.

Since http2 is bundled with Felix Http, which is in the same order as websockets, adding websockets to the bundle as well made most sense to me. Also, since it was part of the former Felix http bundle for Jetty 9/11.

http/jetty12/pom.xml
```diff
@@ -414,6 +422,21 @@
...
+           <groupId>org.eclipse.jetty.websocket</groupId>
+           <artifactId>jetty-websocket-jetty-server</artifactId>
+           <version>${jetty.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.eclipse.jetty.websocket</groupId>
+          <artifactId>jetty-websocket-jetty-client</artifactId>
+          <version>${jetty.version}</version>
+        </dependency>
         <dependency>
```
cziegeler, line 440: I think we should not include the client and only server parts

http/jetty12/pom.xml
```diff
@@ -37,12 +37,12 @@
...
         <developerConnection>scm:git:https://github.com/apache/felix-dev.git</developerConnection>
         <url>https://gitbox.apache.org/repos/asf?p=felix-dev.git</url>
-        <tag>org.apache.felix.http.jetty12-1.0.2</tag>
+        <tag>org.apache.felix.http.jetty12-1.0.3</tag>
     </scm>
 
     <properties>
         <felix.java.version>17</felix.java.version>
-        <jetty.version>12.0.5</jetty.version>
+        <jetty.version>12.0.6</jetty.version>
```
cziegeler, line 45: I've updated to 12.0.6 with a separate issue

cziegeler:
Ok, fair enough. As commented we should probably not include the client. Is there any demo project somewhere that allows to actually test this?

paulrutter:
@cziegeler thanks for the review, i will process the comments and see if i can add a example WebServlet in https://github.com/apache/felix-dev/tree/master/http/samples/whiteboard/src/main/java/org/apache/felix/http/samples/whiteboard as part of this PR.

paulrutter:
I'm having a hard time getting an example to work when using the https://github.com/apache/felix-dev/tree/master/http/samples/whiteboard/src/main/java/org/apache/felix/http/samples/whiteboard bundle, somehow the websocket servlet is not getting initialized. I probably won't get to this part today.

http/jetty12/pom.xml
paulrutter: Done

stbischof:
There is a osgi spec planed for websockets. May be interesting for you to look into this process and be a part of it

paulrutter:
I noticed that and linked it already, would be interesting indeed.

stbischof:
Then please come to the next osgi apec Meeting.

All Public and you are welcome.
https://calendar.google.com/calendar/embed?src=c_fh3lhb5p0l29f6phu2ndifh4a4%40group.calendar.google.com&ctz=America%2FNew_York

paulrutter:
Thanks, i might join depending on my availabity. I added some comments to the draft here https://github.com/osgi/osgi/issues/575#issuecomment-2027711315.

stbischof:
Jetty 12.0.8 is available.

paulrutter:
@cziegeler what would be needed to get this merged?
I haven't gotten to the example yet though, is that the only thing still pending?

paulrutter:
I now have a working example, which outputs something like the following:

```
g! ## [servlet1] Init with config [org.apache.felix.http.base.internal.handler.ServletConfigImpl@3d36ae4f]
## [servlet2] Init with config [org.apache.felix.http.base.internal.handler.ServletConfigImpl@46f7eff4]
## [filter1] Init with config [org.apache.felix.http.base.internal.handler.FilterConfigImpl@3d13614c]
## [filter2] Init with config [org.apache.felix.http.base.internal.handler.FilterConfigImpl@52eac941]
## [websocketservlet1] Init, delaying init call until service method is called...
## [filter1] Filter request [ServletRequestWrapper->org.eclipse.jetty.ee10.servlet.ServletApiRequest@6a131dfe]
## [filter1] Filter request [ServletRequestWrapper->org.eclipse.jetty.ee10.servlet.ServletApiRequest@36faee65]
## [websocketservlet1] Delayed init...
## [websocketservlet1] Ensuring websockets
## [websocketservlet1] Configuring WebSocket factory
## [websocketservlet1] Upgrading to websocket request...
## [websocketservlet1] Done, upgrading to websocket request.
## [class org.apache.felix.http.samples.whiteboard.TestWebSocketServlet$TestWebSocket] Opened session: WebSocketSession[to=PT30S,WebSocketCoreSession@3dc5da96{SERVER,WebSocketSessionState@3406391d{CONNECTED,i=NO-OP,o=NO-OP,c=null},[ws://localhost:8080/filtersample/websocket/test,null,false.[permessage-deflate]],af=true,i/o=4096/4096,fs=65536}->JettyWebSocketFrameHandler@1b5be01d[org.apache.felix.http.samples.whiteboard.TestWebSocketServlet$TestWebSocket],JettyWebSocketFrameHandler@1b5be01d[org.apache.felix.http.samples.whiteboard.TestWebSocketServlet$TestWebSocket]]
## [class org.apache.felix.http.samples.whiteboard.TestWebSocketServlet$TestWebSocket] Received message: test
```

The last message is the websocket message received via the browser, which can be invoked via the Chrome Console:

```
const websocket = websocket = new WebSocket("ws://localhost:8080/filtersample/websocket/test");
websocket.send("test");
```

cziegeler:
Thanks @paulrutter for your work, I think we should go with what we have here and leave the abstract class out of the main bundle for now. Someone who wants to do websockets can easily pick up that class from the sample

paulrutter:
Thanks for merging! I might add some information to the README about this in a separate PR.

Will a new release follow soon?

paulrutter:
Minor doc update pr here https://github.com/apache/felix-dev/pull/307

PR #301, opened 2024-03-05
Update felix-resolver and add github validation

laeubi:
Currently the felix-resolver project requires to be build on a true java 8 / 6 JDK as there is no release support for java 6, it also uses some older OSGi framework dependencies.

This now does the follwoing:
- increase from Java 1.6 > 1.8
- update to the latest OSGi dependencies
- set the version to 3.0.0-SNAPSHOT to account for upgrade of Java+OSGi
- add a github verification that builds at laest the resolver part.

@tjwatson @stbischof please review

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,23 @@
+name: Java CI with Maven
+
+on:
+  push:
+    branches: [ "master" ]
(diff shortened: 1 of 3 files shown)


stbischof:
merged #321 can you rebase and adapt?

PR #303, opened 2024-03-25
Update Felix HTTP documentation

paulrutter:
See https://issues.apache.org/jira/browse/FELIX-6694.
A short summary of the changes made:

- Removed cometd, last released version is from 2014 and nowadays, websockets should be used instead
- update bundle versions
- add jetty 12 and wrappers bundles
- update jetty properties
- update code examples
- update servlet, filter properties. Add OSGi reference links for full documentation.

I would need some help on the HTTP/2 part, as i'm not sure that information is still up-to-date / required.
The Jetty12 bundle seems to offer all relevant HTTP/2 dependencies already. That part of the documentation should be updated still.

Also, when https://github.com/apache/felix-dev/pull/298 is merged, the readme (and ideally the code examples) should be updated to reflect.

@cziegeler for visibility.


paulrutter:
I looked into enabling HTTP/2 support with the Jetty12 bundle, and make some changes to the documentation to reflect this.
HTTP/2 is now up and running for me with Jetty 12.

BTW: It seems that `jetty-alpn-server` makes no sense to have in the Jetty12 bundle (as dependency), as it only works when having that bundle started separatly after the ServiceLoader bundles have been started.

paulrutter:
Then i would first need to know what the difference is with the full bundle; from the looks of it it's just exposing less package in the pom.xml, that's it.
I can add some information on this if that would be helpful though. Do you know what the light version is used for?

paulrutter:
@stbischof i'm not in contact, but i see how this could work.
Although i doubt this should be documented as the bundle is not even published to maven central, i will add a short description for it in the README.

paulrutter:
Ah thanks, i will update the docs to reflect.

http/README.md
```diff
@@ -58,48 +77,75 @@ of additional bundles to deploy [as described in the jetty documentation](https:
...
 ```
+An implementation note for when using the Felix HTTP Jetty 12 bundle: only registering a filter without a servlet, will not work. Make sure that there is a servlet registered (on the same path as the filter) when registering a filter, even when that servlet is not hit eventually.
 
 To ensure the HTTP whiteboard service picks up your servlet and filter correctly, your service
 registration *must* provide several service properties.
 
 
 ### Servlet service properties
 
+See full reference in the [OSGi specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html#service.http.whiteboard-i21223311).
```
stbischof, line 118: https://docs.osgi.org/specification/osgi.cmpn/8.1.0/

we should look to 8.1 release

http/README.md
```diff
@@ -58,48 +77,75 @@ of additional bundles to deploy [as described in the jetty documentation](https:
...
   * `filter.init.*` - these properties (sans the `filter.init.` prefix) are made available throught the `FilterConfig` object of your filter. This allows you to supply your filter initialization parameters as you would normally do in the web descriptor (web.xml).
 
+The order of filters is no longer managed by the `service.ranking` property, but by implementing the `compareTo` method. 
+Multiple servlet filters can process the same servlet request/response. If more than one Filter matches, they are processed in ranking order, as specified in `ServiceReference.compareTo`. The servlet filter with the highest ranking is processed first in the filter chain, while the servlet filter with the lowest ranking is processed last, before the Servlet.service method is called. After the servlet completes its service method the filter chain is unwound in reverse order.
+
 ### ServletContextHelper service properties
 
+See full reference in the [OSGi specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html#service.http.whiteboard.servletcontext).
```
stbischof, line 143: https://docs.osgi.org/specification/osgi.cmpn/8.1.0/

we should look to 8.1 release

stbischof:
LGTM,

When you are ready you may should ask Carsten. 
I would not merge into things where he is the guru

paulrutter:
@cziegeler can you review the changes to see if it makes sense and is correct?

cziegeler (approved):

cziegeler:
Thanks @paulrutter - I applied your changes

PR #305, opened 2024-04-10
FELIX-6688 - Context path of outer servlet container is not respected…

sagarmiglani:
… for authentication

### DIFF
--- a/webconsole/src/main/java/org/apache/felix/webconsole/internal/servlet/OsgiManagerHttpContext.java
+++ b/webconsole/src/main/java/org/apache/felix/webconsole/internal/servlet/OsgiManagerHttpContext.java
@@ -44,13 +44,16 @@ final class OsgiManagerHttpContext extends ServletContextHelper {
 
     private final Bundle bundle;
 
+    private final String webManagerRoot;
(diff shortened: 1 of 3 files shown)


sagarmiglani:
@cziegeler Can you please have a look?

webconsole/src/main/java/org/apache/felix/webconsole/internal/servlet/OsgiManagerHttpContext.java
```diff
@@ -63,19 +66,23 @@ public URL getResource(final String name) {
...
             public String getContextPath() {
-                return "";
+                int managerRootIndex = r.getContextPath().lastIndexOf(webManagerRoot);
+                return r.getContextPath().substring(0, managerRootIndex);
             }
 
             @Override
             public String getServletPath() {
-                return r.getContextPath();
+                int managerRootIndex = r.getContextPath().indexOf(webManagerRoot);
```
cziegeler, line 84: For consistency I think we should use lastIndexOf here as well

cziegeler (approved):

PR #309, opened 2024-04-23
FELIX-6692 Add Jetty WebSocket support for jetty 11.x

enapps-enorman:
I was trying to make this work with jetty 11.x before I saw there was similar work for the jetty12 branch.   So this is a slightly different approach than what was merged for #298 

The changes I'm proposing here will initialize the websocket support in the main ServletContextHandler

This was trying to follow the guidance from the jetty documentation at: https://eclipse.dev/jetty/documentation/jetty-11/programming-guide/index.html#pg-server-websocket

How this changes the http.jetty bundle:

1. No new jetty classes embedded in the http.jetty bundle.  The websocket related classes remain optional and would be expected to be supplied by deploying the appropriate jetty bundles.

2. The following new entries are now in the Import-Package instruction to declare the new optional Import-Package values:

    org.eclipse.jetty.websocket.server.config;version="[11.0,12)";resolution:=optional
    org.eclipse.jetty.websocket.jakarta.server.config;version="[11.0,12)";resolution:=optional

3. These two new configuration properties are available in the JettyConfig that specify whether to enable the two possible websocket implementations

    `org.apache.felix.jakarta.ee9.websocket.enable`

    when set to true, the JakartaWebSocketServletContainerInitializer is called to setup the ServerContainer for the web application context. This also requires that the org.eclipse.jetty.websocket:websocket-jakarta-server bundle is also deployed and active. If that bundle is missing a warning is logged and the websocket initialization is skipped.

    `org.apache.felix.jetty.websocket.enable`
when set to true, the JettyWebSocketServletContainerInitializer is called to setup the JettyWebSocketServerContainer for this web application context. This also requires that the org.eclipse.jetty.websocket:websocket-jetty-server bundle is also deployed and active. If that bundle is missing a warning is logged and the websocket initialization is skipped.

Then one way to declare a custom websocket from a servlet component in a bundle would be to override the servlet init method and do something like below:

    @Override
    public void init(ServletConfig config) throws ServletException {
        super.init(config);

        // The websocket container is attached to the main servlet context, not the per-bundle one.
        //  Lookup the ServletContext for the context path where the websocket server is attached.
        ServletContext servletContext = config.getServletContext();
        ServletContext rootContext = servletContext.getContext("/");

        // Retrieve the JettyWebSocketServerContainer.
        JettyWebSocketServerContainer container = JettyWebSocketServerContainer.getContainer(rootContext);

        // register the mapping to our custom websocket
        container.addMapping("/mywebsocketpath1", (upgradeRequest, upgradeResponse) -> new MyWebSocket1());
    }

### DIFF
--- a/http/base/src/main/java/org/apache/felix/http/base/internal/HttpServiceController.java
+++ b/http/base/src/main/java/org/apache/felix/http/base/internal/HttpServiceController.java
@@ -139,6 +139,15 @@ public final class HttpServiceController
     }
 
     /**
+     * Stores an attribute in the to be created shared servlet context.
+     * @param key attribute key
+     * @param value attribute value
+     */
+    public void setAttributeSharedServletContext(String key, Object value) {
+        this.whiteboardManager.setAttributeSharedServletContext(key, value);
+    }
+
+    /**
      * Stops the http and http whiteboard service.
      */
     public void unregister()
--- a/http/base/src/main/java/org/apache/felix/http/base/internal/whiteboard/WhiteboardManager.java
+++ b/http/base/src/main/java/org/apache/felix/http/base/internal/whiteboard/WhiteboardManager.java
@@ -70,6 +70,7 @@ import org.apache.felix.http.base.internal.whiteboard.tracker.ResourceTracker;
 import org.apache.felix.http.base.internal.whiteboard.tracker.ServletContextHelperTracker;
 import org.apache.felix.http.base.internal.whiteboard.tracker.ServletTracker;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Constants;
 import org.osgi.framework.Filter;
@@ -118,6 +119,7 @@ public final class WhiteboardManager
     private final FailureStateHandler failureStateHandler = new FailureStateHandler();
 
     private volatile ServletContext webContext;
+    private volatile Map<String, Object> attributesForSharedContext = new HashMap<>();
 
     /**
      * Create a new whiteboard http manager
@@ -203,6 +205,7 @@ public final class WhiteboardManager
         this.contextMap.clear();
         this.servicesMap.clear();
         this.failureStateHandler.clear();
+        this.attributesForSharedContext.clear();
         this.registry.reset();
     }
 
@@ -366,6 +369,8 @@ public final class WhiteboardManager
                         {
                             handlerList.add(handler);
                             Collections.sort(handlerList);
+                            setAttributes(handler.getSharedContext());
+
                             this.contextMap.put(info.getName(), handlerList);
 
                             // check for deactivate
@@ -400,6 +405,21 @@ public final class WhiteboardManager
             return true;
         }
         return false;
+    }
+
+    /**
+     * Set the stored attributes on the shared servlet context.
+     * @param context the shared servlet context
+     */
+    private void setAttributes(@Nullable ServletContext context) {
+        if (context != null) {
+            attributesForSharedContext.forEach((key, value) -> {
+                if (key != null && value != null) {
+                    SystemLogger.LOGGER.info("Shared context found, setting stored attribute key: '{}', value: '{}'", key, value);
+                    context.setAttribute(key, value);
+                }
+            });
+        }
     }
 
     /**
@@ -952,5 +972,15 @@ public final class WhiteboardManager
     private void updateRuntimeChangeCount()
     {
         this.serviceRuntime.updateChangeCount();
+    }
+
+    /**
+     * Stores an attribute in the to be created shared servlet context.
+     * @param key attribute key
+     * @param value attribute value
+     */
+    public void setAttributeSharedServletContext(String key, Object value) {
+        SystemLogger.LOGGER.info("Storing attribute for shared servlet context. Key '{}', value: '{}'", key, value);
+        this.attributesForSharedContext.put(key, value);
     }
 }
--- a/http/jetty/pom.xml
+++ b/http/jetty/pom.xml
@@ -44,6 +44,9 @@
         <felix.java.version>11</felix.java.version>
         <jetty.version>11.0.20</jetty.version>
         <baseline.skip>true</baseline.skip>
+        <org.ops4j.pax.exam.version>4.13.3</org.ops4j.pax.exam.version>
+        <!-- To debug the pax process, override this with -D -->
+        <pax.vm.options>-Xmx512M</pax.vm.options>
     </properties>
 
     <build>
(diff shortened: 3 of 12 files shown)


paulrutter:
This is a very nice solution. Would be worthwhile to see if this also works on 12, as this code is a lot neater in terms of initialization.

Can you list the jetty bundles that are required to be deployed in the OSGi environment?

enapps-enorman:
@paulrutter Thanks for reviewing and providing feedback.  I haven't tried this with jetty12, but the documentation looks similar with 2 additional implementations (Jakarta EE 8 and Jakarta EE 10).

After reviewing the jetty12 documentation, I will rename the "org.apache.felix.jakarta.websocket.enable" option to "org.apache.felix.jakarta.ee9.websocket.enable" to avoid conflicts if we decide to go with this solution and expand it to support the other two options for jetty12.

Regarding the jetty bundles that are required, below is what I believe is the minimum set of additional bundles to satisfy the dependencies:

For the jetty specific websocket implementation
---
- org.eclipse.jetty:jetty-webapp
- org.eclipse.jetty.websocket:websocket-core-common
- org.eclipse.jetty.websocket:websocket-core-server
- org.eclipse.jetty.websocket:websocket-jetty-api
- org.eclipse.jetty.websocket:websocket-jetty-common
- org.eclipse.jetty.websocket:websocket-jetty-server
- org.eclipse.jetty.websocket:websocket-servlet
- org.eclipse.jetty:jetty-xml


For the jakarta EE9 specific websocket implementation
---
- jakarta.websocket:jakarta.websocket-api
- org.eclipse.jetty:jetty-alpn-client
- org.eclipse.jetty:jetty-client
- org.eclipse.jetty:jetty-webapp
- org.eclipse.jetty.websocket:websocket-core-client
- org.eclipse.jetty.websocket:websocket-core-common
- org.eclipse.jetty.websocket:websocket-core-server
- org.eclipse.jetty.websocket:websocket-jakarta-client
- org.eclipse.jetty.websocket:websocket-jakarta-common
- org.eclipse.jetty.websocket:websocket-jakarta-server
- org.eclipse.jetty.websocket:websocket-servlet
- org.eclipse.jetty:jetty-xml

paulrutter:
@enapps-enorman could you add an example to the whiteboard sample project? It would help understand the solution and annotations needed.

enapps-enorman:
@paulrutter With my last commit [f256aab](https://github.com/apache/felix-dev/pull/309/commits/f256aabeeddf97af78a835e0fccff489d58e404b) I added some paxexam integration tests to verify the functionality.  

Is that a good enough sample to demonstrate how to use it?

paulrutter:
It would be interesting to see if this approach would also work for the jetty12 bundle, although i'm not a fan of needing to install all those jetty bundles in the OSGi runtime. But that's a matter of choice i suppose.

enapps-enorman:
> It would be interesting to see if this approach would also work for the jetty12 bundle, although i'm not a fan of needing to install all those jetty bundles in the OSGi runtime. But that's a matter of taste i suppose.

Yes that could be interesting, but I am not going to spend any time on experimenting with jetty12 until I get a yes or no regarding whether there is any chance that this proposal is going to be accepted or not.  The source code is available here if you want to try something similar on jetty12 and tell us if it worked.

FYI: In my projects, I am using the "light" classifier variant of the http.jetty bundle which doesn't embed the classes from any of the jetty bundles so I am used to declaring all of the individual jetty bundles that I need.  I prefer this so I have the flexibility to switch to a newer release of jetty without waiting for felix to do a new release with the newer version of the jetty classes embedded.

If the consensus is to embed the content of the extra websocket-* bundles into the non-light variant then I wound not care too much either way.  It would just make the binary a bit bigger.

paulrutter:
Yes, i understand. I think @cziegeler is the one that works on the http part regularly. He can give some more guidance maybe.

cziegeler:
Thanks @enapps-enorman , I like this approach as it is not enlarging the felix http jetty bundle.
I think, it would be good if we have the same approach for jetty and jetty12 to make the transition easier.
@paulrutter Could you maybe try if that works for jetty12 as well?
For the list of additional bundles needed, I don't think that this is really problematic that you need to install all of them in addition. You just have to define that list once.
Now if we think that this is problematic, we could create two additional artifacts which combine all those bundles into a single one, so we would have a felix http jetty bundle with a classifier jettywebsocket and ee9websocket (or whatever names we choose). But we can wait with that one, if too many people complain

paulrutter:
I will try it out @cziegeler, maybe we should wait witg releasing a new version of jetty12 until we know it's feasible or not.

paulrutter:
@enapps-enorman @cziegeler what do you think of this? https://github.com/apache/felix-dev/pull/310

This way, both styles are supported:
* using the main jar won't have websocket classes
* if someone wants to use Jakarta or Jetty WebSocket classes, those classifiers can be used
* if someone wants to run with their own Jetty jars, use light classifier

The PR doesn't contain tests yet, but if we align on the approach, i could do that based on the work from @enapps-enorman.

paulrutter:
If we go along with this PR and #310, i think the following changes would need to be made to this PR:

* Rename `org.apache.felix.jetty.websocket.enable` to `org.apache.felix.jetty.ee9.websocket.enable`, as Jetty12 can support both EE8/9/10, but one needs to specify which EE version to use.
* Update the README to document the new options for the Jetty 11 bundle (see https://github.com/apache/felix-dev/pull/310/files#diff-8e568763d7d4e15f9465300afa93c121c7f8e399acfff4c2aee7e70ed1199401)

enapps-enorman:
@paulrutter Thanks for working out the jetty12 stuff.  I've made an attempt to integrate the code changes you suggested.

paulrutter (approved):

paulrutter:
The changes look good, only left would be to update the documentation to include the new settings.

Update the README to document the new options for the Jetty 11 bundle (see https://github.com/apache/felix-dev/pull/310/files#diff-8e568763d7d4e15f9465300afa93c121c7f8e399acfff4c2aee7e70ed1199401)

enapps-enorman:
> Update the README to document the new options for the Jetty 11 bundle

I could do that but I guess I was thinking that it might be an awkward to merge with two PRs making changes to that file.  I believe at this point the changes in #309 and #310 are changing different files (or identical changes to the same file).  Would it be easier to do a followup PR for the README after the other two are merged?

paulrutter:
We could do that, or i merge in your branch into mine and make the changes there?
Then we have one PR with all the changes for both Jetty bundles.

cziegeler (approved):

paulrutter:
Created PR: https://github.com/apache/felix-dev/pull/312

paulrutter:
@enapps-enorman i just noticed that in the Jetty11 bundle, there is this import: https://github.com/apache/felix-dev/blob/master/http/jetty/src/main/java/org/apache/felix/http/jetty/internal/JettyService.java#L56C1-L56C73

I think this could cause issues when the Jetty websocket bundle is not loaded. Does this need adjusting before the release?

PR #310, opened 2024-04-26
FELIX-6692 jetty 12.x websockets new approach (based on #309)

paulrutter:
Based on the Jetty 11.x websocket work in https://github.com/apache/felix-dev/pull/309, a new PR that incorporates a similar style for Jetty 12.
It adds two new classifiers:

- `with-jetty-ee10-websockets`: jar with the additional Jetty EE10 websocket classes
- `with-jakarta-ee10-websockets`: jar with the additional Jakarta EE10 websocket classes

The main jar and the light jar remain the same.
Two new configuration options:

`org.apache.felix.jakarta.ee10.websocket.enable`
`org.apache.felix.jetty.ee10.websocket.enable` 

Needs additional tests/examples though, not a finished PR yet.


cziegeler:
As we use the same Apache Felix code base for both, maybe this is something that changed in Jetty 12? Although that sounds a little bit wired

paulrutter:
I do see https://github.com/jetty/jetty.project/issues/9925, but that has already been fixed. Maybe i can debug the jetty code a bit to seeel where it fails.

paulrutter:
I found out that https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L510 is the culprit, and setting `context.setCrossContextDispatchSupported(true);` in `JettyService` will solve it.

Now trying if i can get the example code to work.

paulrutter:
```
## [class org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket] Opened session: WebSocketSession[to=PT30S,WebSocketCoreSession@7946968{SERVER,WebSocketSessionState@7eddbcb3{CONNECTED,i=NO-OP,o=NO-OP,c=null},[ws://localhost:8080/filtersample/websocket1/example,null,false.[]],af=true,i/o=4096/4096,fs=65536}->JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket],JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket]]
## [class org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket] Received message: abc
## [class org.apache.felix.http.samples.whiteboard.TestWebSocketServletMainContext$TestWebSocket] Opened session: WebSocketSession[to=PT30S,WebSocketCoreSession@691aa27d{SERVER,WebSocketSessionState@73f6ef48{CONNECTED,i=NO-OP,o=NO-OP,c=null},[ws://localhost:8080/websocket/example,null,false.[]],af=true,i/o=4096/4096,fs=65536}->JettyWebSocketFrameHandler@11f83f04[org.apache.felix.http.samples.whiteboard.TestWebSocketServletMainContext$TestWebSocket],JettyWebSocketFrameHandler@11f83f04[org.apache.felix.http.samples.whiteboard.TestWebSocketServletMainContext$TestWebSocket]]
## [class org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket] Error on session: WebSocketSession[to=PT30S,WebSocketCoreSession@7946968{SERVER,WebSocketSessionState@7eddbcb3{CLOSED,i=NO-OP,o=NO-OP,c={1001=SHUTDOWN,Connection Idle Timeout}},[ws://localhost:8080/filtersample/websocket1/example,null,false.[]],af=true,i/o=4096/4096,fs=65536}->JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket],JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket]] - org.eclipse.jetty.websocket.api.exceptions.WebSocketTimeoutException: Connection Idle Timeout
## [class org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket] Closed session: WebSocketSession[to=PT30S,WebSocketCoreSession@7946968{SERVER,WebSocketSessionState@7eddbcb3{CLOSED,i=NO-OP,o=NO-OP,c={1001=SHUTDOWN,Connection Idle Timeout}},[ws://localhost:8080/filtersample/websocket1/example,null,false.[]],af=true,i/o=4096/4096,fs=65536}->JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket],JettyWebSocketFrameHandler@4e6fb6d0[org.apache.felix.http.samples.whiteboard.TestWebsocketServletBundleContext$TestWebSocket]] - 1001 - Connection Idle Timeout
## [class org.apache.felix.http.samples.whiteboard.TestWebSocketServletMainContext$TestWebSocket] Received message: abc
```

Both examples work, i will push the changes and move on to the unit tests.

paulrutter:
@enapps-enorman @cziegeler I took the unit tests from #309 and adjusted these to run on Jetty12.
All succeed now.

I think we have a solid solution in place now. Let me know what you think.

http/jetty12/src/main/java/org/apache/felix/http/jetty/internal/JettyService.java
```diff
@@ -253,9 +253,12 @@ private void initializeJetty() throws Exception
             loginService.setUserStore(new UserStore());
             this.server.addBean(loginService);
 
-            ServletContextHandler context = new ServletContextHandler(this.config.getContextPath(),
+            ServletContextHandler context = new ServletContextHandler(this.config.getContextPath(),                    
                     ServletContextHandler.SESSIONS);
 
+            // To support WebSockets via https://github.com/apache/felix-dev/pull/310
+            context.setCrossContextDispatchSupported(true);
```
enapps-enorman, line 260: I wonder if there is some way that the (JETTY_WEBSOCKET_CONTAINER_ATTRIBUTE and/or ServerContainer.class.getName()) request attribute could be copied into to (or re-routed from) the  SharedServletContextImpl object that is the delegatee for all of the PerBundleServletContextImpl objects.

If that is possible and works then we wouldn't have to mess with the cross context dispatch flag.  I am not sure what the implications or side effects of that might be.  Plus we would be able to lookup the websocket container directly from the PerBundleServletContextImpl and not have to switch to the root context which seems like an implementation detail that the downstream developers shouldn't need to know.

paulrutter: From what i can find, cross context support allows for dospatching requests between different Jetty contexts, as can be seen in https://github.com/jetty/jetty.project/issues/10617.

For Felix HTTP there is only one context, so i would assume there are no side-effects, but i'm not sure of this.

I agree that having a proper alternative would be better though.

paulrutter: @enapps-enorman i fixed it like this, what do you think?
https://github.com/apache/felix-dev/pull/310/commits/38b99d7b799a3712374d5f5b3608359bb3d57626

I will work on adapting the example code as well.

paulrutter:
I kept both examples in the whiteboard bundle, as both have value:

* https://github.com/apache/felix-dev/pull/310/files#diff-a493047a289b4bd7db717df5c308a98988f0417824efa8819cf0e415a2dc750e is the one that uses the WS container to register the mapping on. 
* https://github.com/apache/felix-dev/pull/310/files#diff-53fea368c43e981c1e37100c78de7e8b2087eb9aa51ceedefdddc6138ae88ca2 is the example that extends `JettyWebSocketServlet`, but requires a workaround (see base class).

For my own project, the first approach wouldn't work, as we register multiple instances of the same servlet class which would then overwrite the WS mapping on the WebSocket container. Using the `JettyWebSocketServlet` approach, this does not happen, as `JettyWebSocketServlet` is used to upgrade a regular `HttpServletRequest` to a WebSocket request.

http/jetty12/src/test/java/org/apache/felix/http/jetty/it/JettySpecificWebsocketIT.java
```diff
@@ -0,0 +1,211 @@
...
+import org.osgi.framework.BundleContext;
+import org.osgi.service.http.HttpService;
+import org.osgi.service.servlet.whiteboard.HttpWhiteboardConstants;
+
+/**
+ *
+ */
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerClass.class)
+public class JettySpecificWebsocketIT extends AbstractJettyTestSupport {
```
enapps-enorman, line 64: Maybe rename this to JettyEE10SpecificWebSocketIT for consistency?

enapps-enorman (approved):
This looks good to me.

http/jetty12/src/test/java/org/apache/felix/http/jetty/it/JettySpecificWebsocketIT.java
paulrutter: Will do

paulrutter:
@cziegeler can you check out this PR and #309? Both are now aligned to use the same approach.
Let me know you agree and this is good to merge for the next HTTP Jetty/Jetty12 release.

cziegeler (approved):

cziegeler:
@paulrutter The additional classifiers look good to me, they should be pushed automatically to maven central when we do the release

PR #311, opened 2024-04-30
FELIX-6699 Skip parsing of repeated Service-Component entries

adamcin:
When a bundle's Service-Component header contains explicit paths and wildcard paths, it is possible for an explicit descriptor entry to be parsed twice, if it is also matched by a wildcard entry. This results in an ERROR level log similar to that described in [FELIX-2325](https://issues.apache.org/jira/browse/FELIX-2325). The use case which requires this mix of wildcard and non-wildcard entries is probably rare, but such is the situation with https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/3241

In this situation, bnd is used to package the host bundle, and so the descriptors for the DS components contained within it are listed explicitly in the Service-Component header value generated by the bnd tool. 

Additionally, the project packages some optional component descriptors in a sidecar fragment bundle which cannot define its own Service-Component header. 

In order for SCR to load the optional components only when the fragment is attached, they must be listed by the host bundle's Service-Component header, but only implicitly by using a wildcard, so that their absence is not logged when the fragment is not attached.

### DIFF
--- a/scr/changelog.txt
+++ b/scr/changelog.txt
@@ -1,11 +1,17 @@
+Changes in 2.2.12
+-----------------
+** Bug
+    * [FELIX-6699] Skip parsing of repeated Service-Component entries
+
 Changes in 2.2.8
+-----------------
 ** Improvement
     * [FELIX-6463] Changing log levels for DS component getState and registration
 
 ** Bug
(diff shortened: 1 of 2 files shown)


tjwatson (approved):
LGTM, thanks!

PR #314, opened 2024-05-03
FELIX-6700 Missing o.o.service.servlet.* export-package in felix jetty

enapps-enorman:
The "light" classifier doesn't export the following packages.  These packages were present in the jar but as a private package.

`org.osgi.service.servlet.context;version="2.0.0";uses:="jakarta.servlet.http,org.osgi.framework"
org.osgi.service.servlet.runtime.dto;version="2.0.0";uses:="org.osgi.dto,org.osgi.framework.dto"
org.osgi.service.servlet.runtime;version="2.0.0";uses:="org.osgi.service.servlet.runtime.dto"
org.osgi.service.servlet.whiteboard;version="2.0.0";uses:="jakarta.servlet"`

This proposal changes those private packages to be exported packages.  Also an integration test is added to verify the functionality.

This ensures that any other bundles that import those packages (i.e. the felix webconsole 5.x) will resolve.

For example, without these changes I get these errors:
```
[ERROR] [bundle-packages] org.apache.felix:org.apache.felix.http.webconsoleplugin:1.1.0: Bundle is importing packages [org.osgi.service.servlet.runtime, org.osgi.service.servlet.runtime.dto] with start order 5 but no bundle is exporting these for that start order.
[ERROR] [bundle-packages] org.apache.felix:org.apache.felix.healthcheck.core:2.2.0: Bundle is importing package org.osgi.service.servlet.context with start order 5 but no bundle is exporting these for that start order.
[ERROR] [bundle-packages] org.apache.felix:org.apache.felix.http.inventoryprinter:1.0.0: Bundle is importing packages [org.osgi.service.servlet.runtime, org.osgi.service.servlet.runtime.dto] with start order 5 but no bundle is exporting these for that start order.
[ERROR] [bundle-packages] org.apache.felix:org.apache.felix.webconsole:5.0.2: Bundle is importing package org.osgi.service.servlet.context with start order 5 but no bundle is exporting these for that start order.
```

### DIFF
--- a/http/jetty/pom.xml
+++ b/http/jetty/pom.xml
@@ -281,7 +281,10 @@
                                     org.osgi.service.http.runtime,
                                     org.osgi.service.http.runtime.dto,
                                     org.osgi.service.http.whiteboard,
-                                    !org.osgi.service.servlet.*,
+                                    org.osgi.service.servlet.context,
+                                    org.osgi.service.servlet.runtime,
+                                    org.osgi.service.servlet.runtime.dto,
+                                    org.osgi.service.servlet.whiteboard,
                                     org.apache.felix.http.jetty,
                                     org.apache.felix.http.javaxwrappers,
                                     org.apache.felix.http.jakartawrappers
@@ -289,7 +292,6 @@
                                 <Private-Package>
                                     org.apache.felix.http.base.*,
                                     org.apache.felix.http.jetty.*,
-                                    org.osgi.service.servlet.*
                                 </Private-Package>
                                 <Import-Package>
                                     org.osgi.service.cm;resolution:=optional;version="[1.3,2)",
--- a/http/jetty/src/test/java/org/apache/felix/http/jetty/it/AbstractJettyTestSupport.java
+++ b/http/jetty/src/test/java/org/apache/felix/http/jetty/it/AbstractJettyTestSupport.java
@@ -153,7 +153,7 @@ public abstract class AbstractJettyTestSupport {
      * @param systemProperty the System property which contains the pathname of the test bundle
      * @return the provisioning option
      */
-    public static UrlProvisionOption testBundle(final String systemProperty) {
+    protected UrlProvisionOption testBundle(final String systemProperty) {
         final String pathname = System.getProperty(systemProperty);
         final File file = new File(pathname);
         return bundle(file.toURI().toString());
--- a/http/jetty/src/test/java/org/apache/felix/http/jetty/it/LightClassifierIT.java
+++ b/http/jetty/src/test/java/org/apache/felix/http/jetty/it/LightClassifierIT.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.felix.http.jetty.it;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.ops4j.pax.exam.CoreOptions.bundle;
+import static org.ops4j.pax.exam.CoreOptions.mavenBundle;
+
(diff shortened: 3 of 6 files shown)


stbischof:
Since this is the light package. I would expect, that is does not contains the o.o.servket package.
Thos you can get from the o.o.servlet api jar.

So no export is needed

enapps-enorman:
> Since this is the light package. I would expect, that is does not contains the o.o.servlet package. you can get this from the o.o.servlet api jar.
> 
> So no export is needed an would prefere remove the o.o.sevlet packages and not export them .

@stbischof  Yes, I suppose you could make the case that the "light" classifier isn't light enough.  But historically the "light" classifier has not been a "minimal" packaging and has always embedded the content of other bundles (see also the embedded org.apache.felix.http.base).  If a more minimal packaging is something that people want, then perhaps creating a new "lightest" classifier without any embedding could be tracked separately in a standalone issue?

The reason I proposed it this way was to be consistent to how it had been done for the previous releases.  From what I can tell. the "light" classifier first arrived with org.apache.felix.http.jetty:4.1.0 and all the later versions I checked have all embedded and exported the org.osgi.service.http.* packages.  

The difference is that org.apache.felix.http.jetty:5.0.0 began embedding and using the new org.osgi.service.servlet.* package as well but it was declared as a private package and not exported.

Instead of having a private copy in org.apache.felix.http.jetty and another copy of the same stuff from the additional org.osgi.service.servlet bundle I proposed and treat those packages the same way the older packages were treated.

I haven't double checked, but if there are two sources for the same org.osgi.service.http.context.* packages then I suspect there could be some classloading boundary troubles when those duplicated packages get resolved to different bundles?

stbischof:
Its right from a view from inside the bundle. There the classes hat been there.

But for bundle-users the classes had not been been exported. So they must bringt the api from Outside.

enapps-enorman:
@stbischof  I'm not sure what that word salad was supposed to mean...

I'll just say again that it doesn't make sense to code it one way for half of the embedded packages and another way for the other half.  Inconsistency leads to confusion and thus the bug report from Antoine.  This PR was simply intended to repair the inconsistency.

stbischof:
What you are doing is a absolute valid way.

Some OSGi Implementations do embedded the osgi api.
And if they embedded it. They should Export it.

What i am try to say is, that users of the light bundle at the moment do not get the classes of o.o.servlet from the felix jetty light bundle.

If my bundles uses the ServletContextHelper, the bundle has a import header in the manifest that describes that. When my bundle starts it will only start if there is an other bundle that Exports the same package.

At the moment felix jetty light is bot the bundle that exports this. You are changing the default behavier.

I would like to have it the way around. No export. And no embedded class. 

In times where developers just used the osgi.cmpn bundle in the setup it may was a good idea to ship also the api with an impl. But now the best practice is to use the api bundle.

Long Story short. Maybe @cziegeler could say what the designed behavier here is.

...
I hope that this fix in which way ever... solves the issue, the webconsole has when it tries to run with the light bundle.  There the ServletContextHelper does not work.

cziegeler (approved):

cziegeler:
The jetty bundle is implementing the OSGi specifications, therefore I think it is good that it also exports those APIs. The point of the light bundle is to allow to easier upgrade used 3rd party libraries. However, updating the OSGi spec API would require an update in the implementation as well.
Long story short, I think this PR is doing the right thing

cziegeler:
Thanks @enapps-enorman

adessaigne:
> The point of the light bundle is to allow to easier upgrade used 3rd party libraries.

Yes, I'm using it to be able to update Jetty as soon as possible when it has a CVE.

Thank you @cziegeler for fixing it.

stbischof:
Thx.

GernotBelger:
> @stbischof Yes, I suppose you could make the case that the "light" classifier isn't light enough. But historically the "light" classifier has not been a "minimal" packaging and has always embedded the content of other bundles (see also the embedded org.apache.felix.http.base). If a more minimal packaging is something that people want, then perhaps creating a new "lightest" classifier without any embedding could be tracked separately in a standalone issue?

I can only vote for that. I would really prefer having a "lightest" package without any other bundles.
Several reasons:
* all other bundles are already available independently, so its easy to just add them (e.g. via maven)
* the current source bundles contain only the sources for the org\apache\felix\http\jetty packages, so if the bundle is fetched from maven, you won't have the sources for base, wrappers etc. Which makes debugging (which I did) a real pain 
* it makes using the bundle in a different ogsi environment harder (what I am trying to do)

I did not found another issue tracker here, so adding this comment to the pull request. Maybe you can give me a hint if there is a better place?

stbischof:
When you are looking for a tracker. It is the apache jira.

GernotBelger:
> I haven't double checked, but if there are two sources for the same org.osgi.service.servlet.* packages then I suspect there could be some classloading boundary troubles when those duplicated packages get resolved to different bundles?

Just stumbled over this comment and want to add: it's not only potential classloader troubles. If you have another source of these packages, what will happen (I encountered it and took me quite some time to understand what happens), is, that the service-tracking might not resolve any more.
Reason beeing is, that you are using the default ServiceTracker#open(false), which then will only resolve services that are directly wired to the corresponding interfaces from the right bundle. So if you add a service that is wired to the other bundle, the service will not be considered a valid source for the service tracker.

PR #316, opened 2024-05-07
webconsole doc: add missing conf props

lucapalano:
[FELIX-6703](https://issues.apache.org/jira/browse/FELIX-6703)


cziegeler (approved):
Thanks!

PR #317, opened 2024-05-07
FELIX-6704: Include config/factory PID in error message

kwin:


kwin:
@cziegeler or @bosschaert Can you review/merge?

PR #318, opened 2024-05-07
FELIX-6705 Upgrade to jetty 12.0.10 

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6705


stbischof:
We may should use a constant for the version in the IT classes .

paulrutter:
Yes, i will adjust that

paulrutter:
Changed to draft, as the jetty12 unit tests fail after moving to 12.0.9.

paulrutter:
I had to add the new https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-ee/12.0.9 bundle to the test.
I also added it to the `Export-Package`.

paulrutter:
Jetty 12.0.10 is incoming, will adjust the PR.
I ran it already with a custom build because 12.0.8 seems to have an issue causing a deadlock in some edge cases.

PR #320, opened 2024-05-08
Idea to add CI pipeline with github actions 

paulrutter:
This unfinished PR shows the idea of using github actions on every PR to the master branch.
It would build all projects in the root pom.xml and executes the unit tests.

Only the http project has been added, itests excluded as these didn’t succeed. That needs more work.

I think this integration could help determine if a PR breaks any existing functionality, has compilation errors or uses outdated snapshot versions.

Example here: https://github.com/blueconic/felix-dev/pull/2


paulrutter:
@enapps-enorman do you have an idea why the integration test doesn't work in GitHub actions? It cannot find the variables set in the pom.xml for the bundle filename.

enapps-enorman:
> @enapps-enorman do you have an idea why the integration test doesn't work in GitHub actions? It cannot find the variables set in the pom.xml for the bundle filename.

@paulrutter Sorry, I don't know why that would fail.  Have you tried configuring it to run maven with the --debug argument to output more details?  Honestly, I don't personally use github actions for testing pull requests.  I use a jenkins organization folder to manage that stuff automatically in a similar way to what the apache sling project does at: https://ci-builds.apache.org/job/Sling/job/modules/

paulrutter:
Ok, thank you. Before diving into this further, i will await a response from the maintainers if this is something that they would like to have going forward.

tjwatson:
I would like to have PR validation with github actions.  Also see #301

laeubi:
@paulrutter what do you think instead of using an aggregator pom (that seems otherwise unused) just using different Jobs, then these can run independently / in parallel.

paulrutter:
I personally like the idea of building all modules, as there might be dependencies between them which you would otherwise not notice when you use different jobs per module. Yes, a build will take longer, but i think there's more value than having separate jobs. Of course, this only goes for modules that are (slightly) dependant, not for entirely related modules.

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup Maven Action
+      uses: s4u/setup-maven-action@v1.7.0
+      with:
+       checkout-fetch-depth: 0
+       java-version: 17
+       java-distribution: temurin
+       maven-version: 3.9.6
+    - name: Build with Maven
```
laeubi, line 31:
```suggestion
    - name: felix-http
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: mvn --debug -Dsurefire.failIfNoSpecifiedTests=false -Dtest=\!MissingWebsocketDependenciesIT* -B clean install
```
laeubi, line 32:
```suggestion
      run: mvn --debug -Dsurefire.failIfNoSpecifiedTests=false -Dtest=\!MissingWebsocketDependenciesIT* -B -f http clean install
```
specify http as the file to build will not require to use a root pom ....

paulrutter: Ah right, like that, yes, that could also work. Then we could drop the root pom indeed.

paulrutter:
When i find the time, i will try to investigate https://github.com/apache/felix-dev/pull/320#issuecomment-2106880844, and remove the root pom.

stbischof:
Also great if we can add the OSGi tck to this test.
@laeubi could you give Support?

laeubi:
I already added the TCKs to Equinox so it should be possible, it just seems at the moment actions are not enabled for this repository at all.

I'll try to provide one PR with a minimal setup and then maybe that can be merged first to base additional thing on top of it.

paulrutter:
I reverted the unneeded changes in my pr and disabled running the integration tests for the http module, somehow those don't run correctly in GitHub actions. Needs more work still.

laeubi:
As Github Actions seem not activate here, you need to test them in "your" fork, go to the settings, enable Actions and then create a PR against your own fork to test them.

paulrutter:
Yes, i already did that. See https://github.com/blueconic/felix-dev/actions/runs/9107381646.

laeubi:
Okay then i misread "somehow those don't run correctly in GitHub actions" sorry for that, I think we should get in the minimal working thing and then build on top of that.

paulrutter:
I meant that i couldn't get the integration tests to run in Github Actions. Hence the `-DskipUTs`.

http/pom.xml
```diff
@@ -42,7 +42,7 @@
         <module>base</module>
         <module>bridge</module>
         <module>inventoryprinter</module>
-        <module>itest</module>
+       <!--<module>itest</module>-->
```
laeubi, line 45: Is this still required now `-DskipITs` is used?

http/samples/whiteboard/pom.xml
```diff
@@ -96,7 +96,7 @@
         <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.http.jetty12</artifactId>
-            <version>1.0.5-SNAPSHOT</version>
+            <version>1.0.7-SNAPSHOT</version>
```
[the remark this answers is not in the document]
paulrutter: Yes, the regular unit tests also don't run with what's in master right now.

http/pom.xml
paulrutter: Yes, the regular unit tests also don't run with what's in master right now.

http/samples/whiteboard/pom.xml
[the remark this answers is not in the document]
paulrutter: I could do without, i will remove it.

stbischof:
GitHub actions are active here. You can use them. Had a discussion with the Apache Infrastruktur team. 

But will only be executed after pr is merged.

Please read also the Policy 

https://infra.apache.org/github-actions-policy.html

https://infra.apache.org/github-actions-secrets.html

paulrutter:
Closing as Github actions is already active.
We would need to start adding more subprojects to it though to be useful.

laeubi:
> We would need to start adding more subprojects to it though to be useful.

My idea would be to everytime one touches a subproject one enables the ci build for that as well.

paulrutter:
Yes, that could work. Although one might pull in a lot of additional work to get things to run properly (i noticed that earlier when running the integration tests for HTTP).

paulrutter:
@laeubi i added Felix HTTP to actions now, see https://github.com/apache/felix-dev/commit/86e5a30b163fd5124ae088e58f68d77085647fb0.

I had to disable one specific IT, as i couldn't find out why it only fails in CI atm.

cziegeler:
@paulrutter  Thats nice! However :) I think we should have separate actions per path. Otherwise we build all configured modules, even if only one has changed. The SCR (and also configadmin) builds take especially long, so its a little bit of a waste to always run them if http changes.

paulrutter:
That's a good point; let me see if we can split that up.

paulrutter:
@cziegeler Fixed in https://github.com/apache/felix-dev/commit/f0e5cc3f01411a4f93c4b6941020f2c97e54c8f4.
There are now separate github action files per subproject.

Maybe not ideal for maintainability, but we can improve on this as we go along.

cziegeler:
Great, thanks!

paulrutter:
We might look into https://resources.github.com/learn/pathways/automation/intermediate/create-reusable-workflows-in-github-actions/ to reduce duplicate lines in the CI files.

laeubi:
@paulrutter with this action you can define several filters, this can then be used to skip some jobs:

https://github.com/dorny/paths-filter

I think that's probably easier to handle than multiple files.

paulrutter:
@laeubi i was searching for something like that indeed. I will look into this later, as i agree that would scale better when we add more projects.

paulrutter:
@laeubi Fixed in https://github.com/apache/felix-dev/commit/f572bd8131df53e1fd39e9655afe6c7758541cc1.

PR #321, opened 2024-05-17
Add github verification for SCR

laeubi:
This adds a github verification workflow for the scr component of felix dev.

I created an example PR how it would looks like after the merge here:
- https://github.com/laeubi/felix-dev/pull/3

FYI @tjwatson @stbischof  I think this should be merged first as it does not require any changes to the code itself to make it run, once we have this we can rebase
- https://github.com/apache/felix-dev/pull/320
- https://github.com/apache/felix-dev/pull/301

on this and further enhance it.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,36 @@
+name: Github CI Build
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+    
+permissions: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up JDK 17
+      uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+        cache: maven
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
+      with:
+        maven-version: 3.9.7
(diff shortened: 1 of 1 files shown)


stbischof:
@tjwatson would it be okay if a merge in felix?

tjwatson:
> @tjwatson would it be okay if a merge in felix?
> 
> I have the rights. But maybe there are some more rules.

It is fine with me to merge.  Please merge if you think it is ready.  Although it will not do anything until we get the repo enabled for actions.

laeubi:
Currently the test parser complains about to large xml data:
 https://github.com/laeubi/felix-dev/pull/3/checks?check_run_id=25087687495

we could either keep this, or I disable the processing of the test files for now or as it only affects the integration test, disable only upload of integration tests...

stbischof:
just upload and no publish

why do you use "stCarolas/setup-maven" and noch the default actions?

paulrutter:
The default actions don't let you specify a maven version.

stbischof:
and what feature has 3.9.6 that we need?

laeubi:
> just upload and no publish
> 
> why do you use "stCarolas/setup-maven" and noch the default actions?

What do you mean by default actions?

One can't tell for sure what the runner have installed (java+maven version) and it might even change or differ between os so for me its good "style" to install wanted java version and wanted maven version.

> and what feature has 3.9.6 that we need?

I don't think there is any certain "feature" just to control what is used as otherwhise it might give surprising results.

laeubi:
> just upload and no publish

done

PR #322, opened 2024-05-25
FELIX-6708 upgrade bnd to 7.0.0

paulrutter:
See https://issues.apache.org/jira/browse/FELIX-6708 and https://github.com/bndtools/bnd/issues/2227#issuecomment-2131313831

To be able to handle Multi-Release jars as dependencies in an OSGi bundle.
This prevents exceptions like:

```
build	24-May-2024 18:40:01	[ERROR] Bundle com.abc:jersey-bundle:bundle:55.0-SNAPSHOT : Classes found in the wrong directory: {META-INF/versions/21/org/glassfish/jersey/innate/VirtualThreadSupport$NonLoomishExecutors.class=org.glassfish.jersey.innate.VirtualThreadSupport$NonLoomishExecutors, META-INF/versions/21/org/glassfish/jersey/innate/virtual/LoomishExecutors.class=org.glassfish.jersey.innate.virtual.LoomishExecutors, META-INF/versions/21/org/glassfish/jersey/innate/VirtualThreadSupport$Java21LoomishExecutors.class=org.glassfish.jersey.innate.VirtualThreadSupport$Java21LoomishExecutors, META-INF/versions/21/org/glassfish/jersey/innate/VirtualThreadSupport.class=org.glassfish.jersey.innate.VirtualThreadSupport}
```


paulrutter:
Probably this code can be removed now as well, as bnd can handle it.

https://github.com/apache/felix-dev/blob/6bb228df51719d495ba26b508fa946f7dab142f7/tools/maven-bundle-plugin/src/main/java/org/apache/felix/bundleplugin/BundlePlugin.java#L2236

paulrutter:
Two integration tests fail still, needs work.

paulrutter:
@hboutemy @rotty3000 any feedback on this pr, as you both worked on upgrading bnd last time it was upgraded.

rotty3000 (approved):
LGTM

paulrutter:
What do you think about the failing Integration tests?

rotty3000:
@paulrutter do this to fix the test:
```patch
diff --git a/tools/maven-bundle-plugin/src/it/no-test-scoped-imports/pom.xml b/tools/maven-bundle-plugin/src/it/no-test-scoped-imports/pom.xml
index 3a15568ced..2eedf2350c 100644
--- a/tools/maven-bundle-plugin/src/it/no-test-scoped-imports/pom.xml
+++ b/tools/maven-bundle-plugin/src/it/no-test-scoped-imports/pom.xml
@@ -53,6 +53,11 @@ under the License.
                 <artifactId>maven-bundle-plugin</artifactId>
                 <version>@project.version@</version>
                 <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <_noimportjava>true</_noimportjava>
+                    </instructions>
+                </configuration>
             </plugin>
         </plugins>
     </build>
```

rotty3000:
In fact, @paulrutter if you want to preserve the original behaviour of maven-bundle-plugin not to import `java.*` you may wish to set that bnd flag default to `true` such that users opt-in by setting it to false.

ref: https://bnd.bndtools.org/instructions/noimportjava.html

laeubi:
I think java packages should simply be always imported, I don't see a reason to disable that, especially as its the default for bnd.

rotty3000:
whatever you prefer! 🤷🏻

paulrutter:
Great, i will adjust the PR and rerun the tests with that instruction.

paulrutter:
The only IT still failing was also failing with bnd 6.3.1.
```
[INFO] --- invoker:3.2.1:run (integration-test) @ maven-bundle-plugin ---
[INFO] Building: archiver\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           archiver\pom.xml ................................. SUCCESS (13.3 s)
[INFO] Building: dep-reduced\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           dep-reduced\pom.xml .............................. SUCCESS (4.9 s)
[INFO] Building: embed-multiple-artifacts\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           embed-multiple-artifacts\pom.xml ................. SUCCESS (4.4 s)
[INFO] Building: existing-metadata-no-update\pom.xml
[INFO] run pre-build script prebuild.groovy
[INFO] run post-build script verify.groovy
[INFO]   The post-build script did not succeed. assert manifest.isEmpty()
       |        |
       |        false
       Manifest-Version: 1.0
       Bnd-LastModified: 1717484827643
       Build-Jdk-Spec: 21
       Bundle-Description: This project has been built before (i.e. a cache fil
        e is there), the current pom.xml has not been touched, nor any of the b
        uilt files
       Bundle-DocURL: https://wcm.io
       Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
       Bundle-ManifestVersion: 2
       Bundle-Name: existing-metadata-no-update
       Bundle-SymbolicName: org.apache.felix.bundleits.existing-metadata-no-upd
        ate
       Bundle-Vendor: wcm.io
       Bundle-Version: 1.0.0.SNAPSHOT
       Created-By: Apache Maven Bundle Plugin 5.1.10-SNAPSHOT
       Tool: Bnd-7.0.0.202310060912
[INFO]           existing-metadata-no-update\pom.xml .............. FAILED (28.2 s)
[INFO] Building: no-test-scoped-imports\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           no-test-scoped-imports\pom.xml ................... SUCCESS (7.9 s)
[INFO] Building: reproducible\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           reproducible\pom.xml ............................. SUCCESS (2.9 s)
[INFO] Building: reproducible-manifest\pom.xml
[INFO] run post-build script verify.groovy
[INFO]           reproducible-manifest\pom.xml .................... SUCCESS (5.7 s)
```

paulrutter:
I think we should update the major version when we release this, due to the minimum jvm version updated to 17.

laeubi:
> I think we should update the major version when we release this, due to the minimum jvm version updated to 17.

bnd major update > felix major update regardless of java version (but plugin minium java version must be set to 17 then)

paulrutter:
Maybe it's platform related? I'm running on a Windows 11 based machine.

rotty3000:
> Maybe it's platform related? I'm running on a Windows 11 based machine.

.. and I'm running on Ubuntu Linux. Bnd has often had weird windows issues because it's the least tests platform.

paulrutter:
If you don't see the failing test, then i'd say we're good to go.

rotty3000:
Perhaps you can confirm that `src/it/existing-metadata-no-update/prebuild.groovy` is actually updating the `lastmodified` time in Windows. (`setLastModified` returns a boolean). It could be a windows file locking thing since the file was just written from src to target for the sake of the tests.

paulrutter:
I printed the return value of `setLastModified` and it returns true:
```
Running pre-build script: C:\git\felix-dev\tools\maven-bundle-plugin\target\it\existing-metadata-no-update\prebuild.groovy
true
```

paulrutter:
I don't have to power to merge though ;) 
And someone should create a release for this, updating the major version as well.

paulrutter:
@tjwatson @cziegeler any feedback on this PR?

paulrutter:
@cziegeler i have been running with 7.0.0 for a whole release cycle now and haven't run into any issues. Is there a chance this can be merged and released?

stbischof:
So we for everyone using felix-bundle-plugin to use java17.

Would update the felix.Framework to java17?

paulrutter:
Since it would be a major version upgrade, i don't see a problem with requiring 17.
Also, since the bundle can be used apart from felix itself, we probably don't need to enforce 17 as a baseline for felix framework itself.

cziegeler:
I agree that updating to bnd 7 and requiring java 17 for the tool chain is probably not a huge problem. However, I'm too far away from the bundle-plugin code to make a call here.

paulrutter:
@rotty3000 maybe?

rotty3000:
man... I've not looked at this for such a long time I would be remise to make a recommendation such out of date info. Perhaps @jbonofre would have better insight?

stbischof:
@paulrutter since this plugin is justr a wrapper around bnd why not using bnd-maven-plugin?

paulrutter:
I looked and `bnd-maven-plugin` and although that would probably work, i find the configuration (in a CDATA tag in the pom.xml) less convenient/readable than the maven-bundle-plugin.
Also, we have a lot of OSGi bundles (80+) which would be quite some rework (with regression risk).

So ideally, i would like to have an updated `maven-bundle-plugin` that uses bnd 7.0.0.

laeubi:
Sad to say but bnd-maven plugin even misses some features of felix-maven-plugin, beside that +1 from me to simply upgrade (with or without major version increment) it all looks fine and sane and should has no impact to the rest of felix.

paulrutter:
Who could give the final go for this to get merged and released? Would be a shame to leave it on the shelve unused, but maybe I'm just to impatient 😉

laeubi (approved):

stbischof (approved):

stbischof:
I can merge, but i am not in the release process.
@jbonofre may knows more.

Sorry for that.

For all other Users i highly recommend to use bnd direct.

laeubi:
> For all other Users i highly recommend to use bnd direct.

Sadly `maven-bundle-plugin` still has some features not supported by `bnd-maven-plugin `and alike directly (or its quite complex to emulate).

chrisrueger:
> Sadly `maven-bundle-plugin` still has some features not supported by `bnd-maven-plugin `and alike directly (or its quite complex to emulate).

Could you list which features these are? That may help to address them in the future.

rotty3000:
`<Embed-Dependency>` is trivially simple in bnd-maven-plugin

https://github.com/bndtools/bnd/blob/b2deb7f91312a4d37f0504d8cfd904847ca028a2/maven-plugins/bnd-maven-plugin/src/it/test-wrapper-bundle/bnd/wrapper.bnd#L7

The embedding can take many forms.

In the case showed, the "maven dependency" `osgi.annotation-*.jar` is embedded into the `lib/` directory (i.e. `lib:=true`) AND is appended to the `Bundle-ClassPath` header (which is created for you automatically with a base of `.` if not already specified so as not to break the bundle's normal class loading structure).

You could embed in a flattened way using  (i.e. the classes and packages are directly beside your own):
```
-includeresource: @osgi.annotation-*.jar
```

and so on... there are lots of possible things you can do like extracting certain files and whatnot!

laeubi:
> <Embed-Dependency> is trivially simple in bnd-maven-plugin

one would think so, but if you

1. don't know it by name
2. don't check in the jar into your code repository

it instantly become incredibly hard. As you mentioned [here](https://stackoverflow.com/questions/58718160/how-to-embed-a-maven-dependency-with-bnd-maven-plugin)

> Sadly, there is no way to specify transitive inclusion natively in bnd

Also  `includeresource` can not filter for scopes (as far as I know) and you have to know the name of the artifact as described [here](https://github.com/bndtools/bnd/issues/3704).

So yes that's all doable but far from convenient to the not so experienced users.

Also please keep in mind this important portion of my answer:

> but that's really cumbersome if **more than one jar is involved**

so all these incredible easy example always ever seem to cover **exactly one jar** where it in fact becomes quite easy.

So to convince people that `bnd-maven-plugin` is a full replacement, I think the best would be to add docs that cover [the examples](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_embedding_dependencies) mentioned here:

```
<!-- embed all compile and runtime scope dependencies -->
<Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>

<!-- embed any dependencies with artifactId junit and scope runtime -->
<Embed-Dependency>junit;scope=runtime</Embed-Dependency>

<!-- inline all non-pom dependencies, except those with scope runtime -->
<Embed-Dependency>*;scope=!runtime;type=!pom;inline=true</Embed-Dependency>

<!-- embed all compile and runtime scope dependencies, except those with artifactIds in the given list -->
<Embed-Dependency>*;scope=compile|runtime;inline=false;artifactId=!cli|lang|runtime|tidy|jsch</Embed-Dependency>

<!-- inline contents of selected folders from all dependencies -->
<Embed-Dependency>*;inline=images/**|icons/**</Embed-Dependency>
```

rotty3000:
Fair, but these are incorrect assumptions:

> don't know it by name
You can use a regex which gets all dependencies? matching is a regex across the names of all artifacts.. so just match all!

> don't check in the jar into your code repository
You don't need to care where the jar is, only that maven will ultimately already have gotten it for you!

The rest may be arguably true, but this is widely used in very large projects with MANY jars (please, if you dare, look at [this repo](https://github.com/liferay/liferay-portal) and search for `-includeresource`).

rotty3000:
anyway.. I do see your point.

laeubi:
> The rest may be arguably true, but this is widely used in very large projects with MANY jars (please, if you dare, look at [this repo](https://github.com/liferay/liferay-portal) and search for `-includeresource`).

And it is using.... [`maven-bundle-plugin`](https://github.com/liferay/liferay-portal/blob/8167e7875f59773e7737b7853a7dbf9c6ddeb406/modules/etl/talend/pom.xml#L87)   ;-)

Don't get me wrong its all possible for sure, it was just asked *why* people using it and I gave some examples, you should not assume that every maven user that wants to wrap a small (or medium) jar can afford to dive into the details of bnd instruction magic, those people only search for a fast way to archive their goal without much hassle and the plugin is offering that for them.

So if these one-liner examples above can be transformed into a one-liner bnd instruction (or better maven configuration option) it is fine and probably there is just a lack of let people know, if I look here:

https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-maven-plugin

I don't find any good hint to archive things similar to `felix-maven-plugin`, (e.g. migration guide) nor how to archive the embedding in a similar way.

cziegeler:
While I really value this discussion and I dont want to stop it, I think this PR is really not the right place. No one will find this here. So please move it to our mailing list - this is the place where discussions really should happen

PR #329, opened 2024-07-10
FELIX-6720 enable virtual thread support in jetty12

paulrutter:
- Add `org.apache.felix.http.jetty.virtualthreads.enable` as configuration option, defaulting to false
- Add logging if virtual threads are used
- Use reflection to detect if Virtual Threads can be used and throw if this is enabled but cannot be used
- Add integration test

I haven't tested it myself yet (apart from the IT), so before merging we should make sure it works properly.


stbischof:
@timothyjward 
May be interesting to what the Performance difference is.

paulrutter:
Having read the webtide blogs, it's not a silver bullet, although it can help in some situations though.

https://webtide.com/if-virtual-threads-are-the-solution-what-is-the-problem/

cziegeler (approved):
lgtm, thanks!

paulrutter:
I've extended the implementation with allowing to set the max number of virtual threads in https://github.com/apache/felix-dev/commit/25bb6d45c85dea45bdaf4c1e895d22f43525a597, as demonstrated in https://jetty.org/docs/jetty/12/programming-guide/arch/threads.html#thread-pool-virtual-threads.

PR #331, opened 2024-08-01
Adds missing filter wrappers

d0mmi:
I added some warappers for the filters that are missing in the current implementation


cziegeler:
Thanks @d0mmi - I created https://issues.apache.org/jira/browse/FELIX-6723 for this.

d0mmi:
@cziegeler Hi, is it known when the next version that will contain this change will be released?

PR #334, opened 2024-09-10
 avoids IllegalArgumentException and memory hoarding

juergen-albert:
fixes https://issues.apache.org/jira/browse/FELIX-6726

ungetServiceObject may be called twice in some cases, which causes an
IllegalArgumentException to be thrown and logged. It furthermore causes
to recreated ComponentServiceObjects to be created and never cleaned up
afterwards.


juergen-albert:
@tjwatson It would be great if you could have a look at this.

cziegeler:
@tjwatson I am not 100% sure but I think we discussed how to handle multiple references to a single prototype from one component in the spec group and decided to handle this as a single reference which is then injected into the various fields/methods.

tjwatson:
> @tjwatson I am not 100% sure but I think we discussed how to handle multiple references to a single prototype from one component in the spec group and decided to handle this as a single reference which is then injected into the various fields/methods.

I see that in https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.component.html#service.component-reference.scope which says:

> prototype_required - For all references to a given bound service, each activated component instance must use a single, distinct service object. That is, for a given bound service, each component instance will use a distinct service object but within a component instance all references to the bound service will use the same service object.

But the problem is then later the spec says this which has some contradiction:

> For a bound service of a reference with prototype or prototype_required reference scope, SCR must use a Service Objects object obtained from the OSGi Framework's service registry using the component's Bundle Context to get any service objects. If service objects for a bound service have been obtained and the service becomes unbound, SCR must unget any unreleased service objects using the Service Objects object obtained from the OSGi Framework's service registry using the component's Bundle Context. This means that if a component instance used a Component Service Objects object to obtain service objects, SCR must track those service objects so that when the service becomes unbound, SCR can unget any unreleased service objects.

The problem is implementing this `unget` of the service objects when the service is "unbound".  What is suppose to happen if the same service is injected into more than one reference for a component but later only one is "unbound" from a dynamic optional reference?  What the implementation does is `unget` all instances of the service object which is the same instance still bound to the other references.  This is what leads to the problem.  Once the other references un "unbound" SCR still has a reference to the injected object and it ungets it again.  If we had instead let each reference get its own managed instance this would not be a problem.  But as it stands now we have this problem where we are ungetting service objects when we shouldn't while there are still other references in the component that still need the service object.

juergen-albert:
The test attached to this PR provokes the Behavior.

tjwatson:
> What you describe seems to be an issue as well, but I'm not sure if this is what is causing this issue here.

It is related though because of this line of code in `invokeUnbindMethod`:

https://github.com/apache/felix-dev/blob/c32fcc418ef1041078131ff63c4e318d26bbafbc/scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java#L2231

That ungets all known service objects that were obtained from the `ServiceObjects` from the framework.  It then does the unget again for the actual reference from the `field` reference.  But the real problem IMO is that it ungets all the known objects when the reference service is "unbound".

tjwatson:
> That ungets all known service objects that were obtained from the `ServiceObjects` from the framework. It then does the unget again for the actual reference from the `field` reference. But the real problem IMO is that it ungets all the known objects when the reference service is "unbound".

What I think needs to happen here is that it should not call `componentContext.getComponentServiceObjectsHelper().closeServiceObjects(refPair.getRef());` until all references for that same `ServiceReference` have been completely unbound.

But this still has weird corner cases if the `ComponentServiceObjects` was injected as a dynamic/optional reference and only that reference gets unbound (because of some target filter).  It would seem like that situation should unget all the services obtained by that `ComponentServiceObjects`.  But currently the implementation shared the same `ComponentServicesObject` instance for every reference in a component instance.

All of this would have been so much more clear if each reference used a different `ComponentServicesObject` to manage their individual instance.  But the spec took a different approach which makes this more confusing and hard to implement.

juergen-albert:
How the hell did I manage to debug this setup the last time I worked on that?

@tjwatson The main issue, is that https://github.com/apache/felix-dev/blob/c32fcc418ef1041078131ff63c4e318d26bbafbc/scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java#L1948
 produces returns a collection with 2 `RefPair` which both represent the same Object. I know this is kind of what you referred to in your issue description above, in my case , I have only one Reference declared, so I would expect only one RefPair to be here.

tjwatson:
> produces returns a collection with 2 `RefPair` which both represent the same Object. I know this is kind of what you referred to in your issue description above, in my case , I have only one Reference declared, so I would expect only one RefPair to be here.

For what it is worth, I don't see that while debugging your test, but I also don't see the actual exception get thrown, instead it is only logged.

juergen-albert:
Strange. I can't confirm or deny what you see, as I'm unable to debug at the Moment. That the Exception is only logged is normal.

tjwatson:
> How the hell did I manage to debug this setup the last time I worked on that?

Don't forget you have to uncomment this line in your test class to debug:
```
        // uncomment to enable debugging of this test class
        //        paxRunnerVmOption = DEBUG_VM_OPTION;
```

juergen-albert:
I finally have the cause of the issue here in this case and I actually think my suggest fix is correct:

1. `DependencyManager.close()` is called for `org.apache.felix.scr.integration.components.felix6726.Consumer` 
2. `RefPair` for  with the Prototype Instance of `org.apache.felix.scr.integration.components.felix6726.BImpl` is found
3. `invokeUnbindMethod` is called in https://github.com/apache/felix-dev/blob/a623362045459e3aef6b166fdb06bb8cacf3c4d0/scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java#L1961
4. `invokeUnbindMethod` already closes and ungets the serivce in https://github.com/apache/felix-dev/blob/a623362045459e3aef6b166fdb06bb8cacf3c4d0/scr/src/main/java/org/apache/felix/scr/impl/manager/DependencyManager.java#L2230
5. then `boundRef.ungetServiceObject` is called afterwards and now the Exception is thrown.

tjwatson:
The more I look at this the more concerned I get that we are not tracking usage of prototype instances correctly for ungetting them.  Rather than spending more time trying to untangle that problem; If we want to proceed with any speed then I suggest we simply catch the IllegalArgumentException in `org.apache.felix.scr.impl.manager.AbstractPrototypeRefPair.doUngetService(ScrComponentContext, T)` and ignore for now.

PR #335, opened 2024-09-30
FELIX-6728 Update to Jetty 12.0.14

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6728


paulrutter:
@cziegeler can we get this one merged? Thanks!

PR #341, opened 2024-10-28
Add locks to BundleWiringImpl.

hq6:
The lack of locking in findClassOrResourceByDelegation and loadClass is causing race conditions that manifest as `NoClassDefFoundError` in a killbill deployment.

This PR is intended to fix these by adding synchronization.


hq6:
This PR will be easier to understand if you hide whitespace changes:
https://github.com/apache/felix-dev/pull/341/files?diff=unified&w=1

PR #356, opened 2024-12-01
FELIX-6741 Upgrade dependencies for maven-bundle-plugin

paulrutter:
- Upgrade all dependencies, including bnd 7.3.0

I will first test this release myself before merging it, as the bnd 7.1.0 mentions a change that might be relevant / no longer backwards compatible. See https://github.com/bndtools/bnd/wiki/Changes-in-7.1.0#bnd

```
Improvement / Fix: Import-Package (coming from Export-Package) are not automatically added to generated Manifest anymore, if the export does not specify a version. This now conforms better to the [specification](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3238802) which requires a version on exports. Notice that this results in different generated Manifests if exports do not specify a version. (PR #6270)
```


paulrutter:
I made some additional changes and added a integration test for https://github.com/bndtools/bnd/pull/6270, to verify impact on the maven-bundle-plugin (there is no impact).

This PR is now ready for review.

PR #359, opened 2024-12-11
Use OSGi Annotations to produce felix.log manifest and clean up.

wilx:
While I was working on reviving felix.logback I was going through the code and I wondered if it could be modernized a little bit. This is the result. I am not sure if it is useful for the project as I have no idea what minimum platform and JVM version it targets.

### DIFF
--- a/log/pom.xml
+++ b/log/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>org.apache.felix</groupId>
     <artifactId>felix-parent</artifactId>
-    <version>6</version>
+    <version>9</version>
     <relativePath>../pom/pom.xml</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
@@ -49,11 +49,42 @@
       <groupId>org.osgi</groupId>
       <artifactId>osgi.core</artifactId>
       <version>6.0.0</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.cm</artifactId>
+      <version>1.6.0</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.service.log</artifactId>
       <version>1.5.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.annotation.bundle</artifactId>
+      <version>1.0.0</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.annotation.versioning</artifactId>
+      <version>1.0.0</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.namespace.service</artifactId>
+      <version>1.0.0</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.component.annotations</artifactId>
+      <version>1.5.1</version>
+      <scope>provided</scope>
     </dependency>
    </dependencies>
   <build>
@@ -61,24 +92,16 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.5.0</version>
+        <version>5.1.9</version>
         <extensions>true</extensions>
         <configuration>
           <instructions>
             <Export-Package>org.osgi.service.log,org.osgi.service.log.admin</Export-Package>
-            <Private-Package>org.apache.felix.log</Private-Package>
-            <Bundle-SymbolicName>${pom.artifactId}</Bundle-SymbolicName>
-            <Bundle-Activator>${pom.artifactId}.Activator</Bundle-Activator>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
             <Bundle-Vendor>The Apache Software Foundation</Bundle-Vendor>
             <Include-Resource>META-INF/LICENSE=LICENSE,META-INF/NOTICE=NOTICE,META-INF/DEPENDENCIES=DEPENDENCIES</Include-Resource>
-            <Require-Capability><![CDATA[
-                osgi.service;filter:="(objectClass=org.osgi.service.cm.ConfigurationAdmin)";effective:=active
-            ]]></Require-Capability>
-            <Provide-Capability><![CDATA[
-                osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.osgi.service.log,org.osgi.service.log.admin",
-                osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.osgi.service.log,org.osgi.service.log.admin",
-                osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.osgi.service.log,org.osgi.service.log.admin"
-            ]]></Provide-Capability>
+            <_reproducible>true</_reproducible>
+            <niceManifest>true</niceManifest>
           </instructions>
         </configuration>
       </plugin>
--- a/log/pom.xml
+++ b/log/pom.xml
@@ -92,7 +92,7 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <version>5.1.9</version>
+        <version>6.0.0</version>
         <extensions>true</extensions>
         <configuration>
           <instructions>
--- a/log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
+++ b/log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
@@ -32,6 +32,7 @@ public class FormatterLoggerImpl extends LoggerImpl implements FormatterLogger {
         super(name, bundle, log, loggerAdmin);
     }
 
+    @Override
     String format(String format, LogParameters logParameters) {
         StringBuilder sb = new StringBuilder();
 
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -8,6 +8,7 @@ on:
       - 'http/**'
       - 'tools/maven-bundle-plugin/**'
       - 'tools/osgicheck-maven-plugin/**'
+      - 'log/**'
       - 'webconsole/**'
   pull_request:
     branches: [ "master" ]
@@ -17,7 +18,8 @@ on:
       - 'tools/maven-bundle-plugin/**'
       - 'tools/osgicheck-maven-plugin/**'
       - 'webconsole/**'
-    
+      - 'log/**'
+
 permissions: {}
 
 jobs:
@@ -50,10 +52,16 @@ jobs:
             - 'tools/osgicheck-maven-plugin/**'
           webconsole:
             - 'webconsole/**'
+          log:
+            - 'log/**'
+
     - name: Felix SCR
       if: steps.changes.outputs.scr == 'true'
       run: mvn -B -V -Dstyle.color=always --file scr/pom.xml clean verify
-    - name: Felix HTTP 
+    - name: Felix Log
+      if: steps.changes.outputs.log == 'true'
+      run: mvn -B -V -Dstyle.color=always --file log/pom.xml clean verify
+    - name: Felix HTTP
       if: steps.changes.outputs.http == 'true'
       run: mvn -B -V -Dstyle.color=always "-Dit.test=!MissingWebsocketDependenciesIT" --file http/pom.xml clean install verify
     - name: Felix Maven bundle plugin 
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Set up Maven
       uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
       with:
-        maven-version: 3.9.7
+        maven-version: 3.9.9
     - name: Check which subproject changed and build affected ones
       uses: dorny/paths-filter@v3
       id: changes
--- a/log/src/main/java/org/apache/felix/log/Activator.java
+++ b/log/src/main/java/org/apache/felix/log/Activator.java
@@ -116,7 +116,7 @@ public final class Activator implements BundleActivator
         String storeDebugPropValue = context.getProperty(STORE_DEBUG_PROPERTY);
         if (storeDebugPropValue != null)
         {
-            storeDebug = Boolean.valueOf(storeDebugPropValue).booleanValue();
+            storeDebug = Boolean.parseBoolean(storeDebugPropValue);
         }
 
         return storeDebug;
--- a/log/src/main/java/org/apache/felix/log/Activator.java
+++ b/log/src/main/java/org/apache/felix/log/Activator.java
@@ -21,10 +21,14 @@ package org.apache.felix.log;
 import java.util.Dictionary;
 import java.util.Hashtable;
 
+import org.osgi.annotation.bundle.Header;
+import org.osgi.annotation.bundle.Requirement;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Constants;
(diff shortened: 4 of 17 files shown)


stbischof:
Just to validate.
Could you show Manifest before and after here as comments.

Updating parent pom seems okay.
Annotations are made dor that. So also good.

Need to look deeper but think this is fine.

log/pom.xml
```diff
@@ -49,6 +49,30 @@
...
+      <groupId>org.osgi</groupId>
+      <artifactId>osgi.annotation</artifactId>
+      <version>8.1.0</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.component.annotations</artifactId>
+      <version>1.5.1</version>
+      <optional>true</optional>
```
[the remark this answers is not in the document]
wilx: It makes it cleaner for users of this artifact. The annotations are only needed during compile time and are then discarded. The `provided` scope shows up in project dependencies for anyone who would use this as a dependency. While `optional` dependencies do not propagate automatically and have to be added by user of this artifact manually. See "[How do optional dependencies work?](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html)" in Maven documentation.

wilx:
> Could you show Manifest before and after here as comments.

It is slightly different. I will provide the comparison later today.

log/pom.xml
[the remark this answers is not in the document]
laeubi: I don't think that optional is correct here. As you mentioned the doc states that optional should be used if there are different providers (e.g. slf4j, jpa, jms,...) and not to hide something (what is shown to the user is then more a decision of the tooling)

laeubi:
> While I was working on reviving felix.logback I was going through the code and I wondered if it could be modernized a little bit.

There is also now some github validation so you should enable that for your module as well, just duplicate the line with your module.

https://github.com/apache/felix-dev/blob/65ac38868a1b09d0dce5ce12a95bcecef7cb8178/.github/workflows/maven.yml#L30

log/pom.xml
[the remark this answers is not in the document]
rotty3000: You should be able to use `<scope>provided</scope>` for this case.

rotty3000 (changes requested):
Just a few nits.

Besides that, thank you @wilx for breathing some life into this bundle. :)

log/pom.xml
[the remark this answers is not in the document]
wilx: > You should be able to use `<scope>provided</scope>` for this case.

I have removed the `optional`. I don't think `provided` scope is appropriate either because there is no other provider of these. Nothing bad will happen if the artifacts stay on classpath.

wilx:
I have pushed some changes.

This is the difference between the 1.3.0 version and the branch snapshot version in MANIFEST.MF after those changes. I have made manually concatenated the MANIFEST.MF lines for easier comparison. There are differences in `uses` for the `Provide-Capability`.

```diff
--- scratch_1.MF        2024-12-13 11:35:27.643249474 +0100
+++ scratch_2.MF        2024-12-13 11:35:55.859447078 +0100
@@ -1,7 +1,5 @@
 Manifest-Version: 1.0
-Bnd-LastModified: 1684133251156
-Build-Jdk: 11.0.19
-Built-By: jbonofre
+Build-Jdk-Spec: 17
 Bundle-Activator: org.apache.felix.log.Activator
 Bundle-Description: A simple implementation of the OSGi R8 Log service
  .
@@ -11,8 +9,8 @@
 Bundle-Name: Apache Felix Log Service
 Bundle-SymbolicName: org.apache.felix.log
 Bundle-Vendor: The Apache Software Foundation
-Bundle-Version: 1.3.0
-Created-By: Apache Maven Bundle Plugin
+Bundle-Version: 1.3.1.SNAPSHOT
+Created-By: Apache Maven Bundle Plugin 5.1.9
 Export-Package: org.osgi.service.log;version="1.5";uses:="org.osgi.fra
  mework",org.osgi.service.log.admin;version="1.0";uses:="org.osgi.serv
  ice.log"
@@ -20,8 +18,9 @@
  k.wiring;version="[1.2,2)",org.osgi.service.log;version="[1.5,1.6)",o
  rg.osgi.service.log.admin;version="[1.0,1.1)",org.osgi.util.tracker;v
  ersion="[1.5,2)"
-Provide-Capability: osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.osgi.service.log,org.osgi.service.log.admin",
- osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.osgi.service.log,org.osgi.service.log.admin",
- osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.osgi.service.log,org.osgi.service.log.admin"
+Provide-Capability: osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.apache.felix.log,org.osgi.service.log",
+ osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.apache.felix.log,org.osgi.service.log,org.osgi.service.log.admin",
+ osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.apache.felix.log,org.osgi.service.log.admin"
 Require-Capability: osgi.service;filter:="(objectClass=org.osgi.service.cm.ConfigurationAdmin)";effective:=active,osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
-Tool: Bnd-3.5.0.201709291849
+Tool: Bnd-6.3.1.202206071316
+
```

rotty3000 (approved):
LGMT (though I do wish we would use `provided` with the "build time" dependencies.)

log/pom.xml
```diff
@@ -49,36 +49,49 @@
...
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.namespace.service</artifactId>
+      <version>1.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.osgi</groupId>
+      <artifactId>org.osgi.service.component.annotations</artifactId>
+      <version>1.5.1</version>
+    </dependency>
```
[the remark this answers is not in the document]
wilx: I kinda disagree but in the end it does not matter if the JAR is there or not, so I have added the provided scope.

log/pom.xml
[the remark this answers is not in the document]
stbischof: The annotations are just relevant on build time for tools loke bnd. If the jar is build the annotation is not visuable at runtime.

Thats why provided is best Option

stbischof:
@laeubi  could you look on the ci?

.github/workflows/maven.yml
```diff
@@ -2,13 +2,15 @@ name: Github CI Build
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "**" ]
```
laeubi, line 5: I would keep master here

laeubi (commented):
I also think the branch spec should only be master at the moment, the rest of the CI change looks fine

.github/workflows/maven.yml
wilx: I have reverted this change.

stbischof:
@wilx could you resolve the merge conflict on ` .github/workflows/maven.yml`?

then i would like to merge

wilx:
> @wilx could you resolve the merge conflict on ` .github/workflows/maven.yml`?
> 
> then i would like to merge

I have rebased the branch on master.

log/pom.xml
```diff
@@ -49,36 +49,52 @@
...
+      <scope>provided</scope>
+    </dependency>
    </dependencies>
   <build>
     <plugins>
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.5.0</version>
+        <version>5.1.9</version>
```
paulrutter, line 89: There is a newer bundle-plugin released in the meantime, using bnd 7.0.0.
It's versioned as 6.0.0.

I've also approved the workflow run, it's running now.

wilx: I have updated the Maven Bundle plugin version.

stbischof:
<dependency> 
       <groupId>org.osgi</groupId> 
       <artifactId>osgi.annotation</artifactId> 
       <version>8.1.0</version> 
       <scope>provided</scope> 
     </dependency>



Could you please use the elementar API dependencies and not the aggregator.

wilx:
@stbischof Done.

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
```diff
@@ -26,12 +26,13 @@
 
 public class FormatterLoggerImpl extends LoggerImpl implements FormatterLogger {
 
-    public FormatterLoggerImpl(
```
stbischof, line 29: This forces an version change. Why should this not be public?

stbischof (changes requested):

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
wilx: Is there an expectation that this would be usable outside the JAR itself? The `org.apache.felix.log` package is not exported at all.

wilx:
The diff with manually concatenated lines for easier comparison:
```diff
`--> diff -upd ./org.apache.felix.log-1.3.0/META-INF/MANIFEST.MF ./org.apache.felix.log-1.3.1-SNAPSHOT/META-INF/MANIFEST.MF
--- ./org.apache.felix.log-1.3.0/META-INF/MANIFEST.MF	2025-03-17 19:53:14.737422780 +0100
+++ ./org.apache.felix.log-1.3.1-SNAPSHOT/META-INF/MANIFEST.MF	2025-03-17 19:52:52.601235604 +0100
@@ -1,7 +1,5 @@
 Manifest-Version: 1.0
-Bnd-LastModified: 1684133251156
-Build-Jdk: 11.0.19
-Built-By: jbonofre
+Build-Jdk-Spec: 17
 Bundle-Activator: org.apache.felix.log.Activator
 Bundle-Description: A simple implementation of the OSGi R8 Log service
  .
@@ -11,10 +9,11 @@ Bundle-ManifestVersion: 2
 Bundle-Name: Apache Felix Log Service
 Bundle-SymbolicName: org.apache.felix.log
 Bundle-Vendor: The Apache Software Foundation
-Bundle-Version: 1.3.0
-Created-By: Apache Maven Bundle Plugin
+Bundle-Version: 1.3.1.SNAPSHOT
+Created-By: Apache Maven Bundle Plugin 6.0.0
 Export-Package: org.osgi.service.log;version="1.5";uses:="org.osgi.framework",org.osgi.service.log.admin;version="1.0";uses:="org.osgi.service.log"
 Import-Package: org.osgi.framework;version="[1.8,2)",org.osgi.framework.wiring;version="[1.2,2)",org.osgi.service.log;version="[1.5,1.6)",org.osgi.service.log.admin;version="[1.0,1.1)",org.osgi.util.tracker;version="[1.5,2)"
-Provide-Capability: osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.osgi.service.log,org.osgi.service.log.admin",osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.osgi.service.log,org.osgi.service.log.admin",osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.osgi.service.log,org.osgi.service.log.admin"
+Provide-Capability: osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.apache.felix.log,org.osgi.service.log",osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.apache.felix.log,org.osgi.service.log,org.osgi.service.log.admin",osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.apache.felix.log,org.osgi.service.log.admin"
 Require-Capability: osgi.service;filter:="(objectClass=org.osgi.service.cm.ConfigurationAdmin)";effective:=active,osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
-Tool: Bnd-3.5.0.201709291849
+Tool: Bnd-7.0.0.202310060912
+
```

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
wilx: I am sorry, I am getting mixed signals here. Do you want me to put the `public` back?

wilx:
> osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.apache.felix.log,org.osgi.service.log"

This is different because what I declared, which is IMHO more accurate:
```java
@Capability(
        namespace = ServiceNamespace.SERVICE_NAMESPACE,
        attribute = { "objectClass:List<String>=\"org.osgi.service.log.LogReaderService\"" },
        uses = { LogReaderServiceImpl.class, LogReaderService.class }
)
final class LogReaderServiceImpl implements LogReaderService
```

> osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.apache.felix.log,org.osgi.service.log,org.osgi.service.log.admin"

```java
@Capability(
        namespace = ServiceNamespace.SERVICE_NAMESPACE,
        attribute = { "objectClass:List<String>=\"org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory\"" },
        uses = { LogServiceImpl.class, LogService.class, LoggerFactory.class, LoggerAdmin.class }
)
final class LogServiceImpl implements LogService
```

> objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.apache.felix.log,org.osgi.service.log.admin"

```java
@Capability(
        namespace = ServiceNamespace.SERVICE_NAMESPACE,
        attribute = { "objectClass:List<String>=\"org.osgi.service.log.admin.LoggerAdmin\"" },
        uses = { LoggerAdminImpl.class, LoggerAdmin.class }
)
public class LoggerAdminImpl implements LoggerAdmin {
```

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
stbischof: i would not change it. 
@laeubi  opinion?

laeubi: If it is not exported then it is effective private in OSGi, in any case I would add such change in a separate PR if desired and do not mix it with the usage of annotations.

Beside that it will only be a micro change.

log/pom.xml
```diff
@@ -49,36 +49,58 @@
...
             <Include-Resource>META-INF/LICENSE=LICENSE,META-INF/NOTICE=NOTICE,META-INF/DEPENDENCIES=DEPENDENCIES</Include-Resource>
-            <Require-Capability><![CDATA[
-                osgi.service;filter:="(objectClass=org.osgi.service.cm.ConfigurationAdmin)";effective:=active
-            ]]></Require-Capability>
-            <Provide-Capability><![CDATA[
-                osgi.service;objectClass:List<String>="org.osgi.service.log.LogReaderService";uses:="org.osgi.service.log,org.osgi.service.log.admin",
-                osgi.service;objectClass:List<String>="org.osgi.service.log.LogService,org.osgi.service.log.LoggerFactory";uses:="org.osgi.service.log,org.osgi.service.log.admin",
-                osgi.service;objectClass:List<String>="org.osgi.service.log.admin.LoggerAdmin";uses:="org.osgi.service.log,org.osgi.service.log.admin"
-            ]]></Provide-Capability>
+            <_reproducible>true</_reproducible>
```
laeubi, line 103:
```suggestion
            <_reproducible>true</_reproducible>
            <niceManifest>true</niceManifest>
```

Can we also enable the nice manifest option?

https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#niceManifest

this makes the generated manifest much easier to read.

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
stbischof: So please let it Public fot this PR

log/pom.xml
stbischof: I am okay with that

wilx: Done.

log/src/main/java/org/apache/felix/log/FormatterLoggerImpl.java
wilx: Reverted the public removal.

stbischof (approved):

stbischof:
Thank you!

PR #362, opened 2024-12-14
Implement 101.8.4 Log Events

wilx:
While I was looking around the felix.log I noticed that the spec section "101.8.4 Log Events" is not implemented in felix.log. So implemented it. Then when I wrote the test, I found out it was actually implemented in felix.eventadmin instead as per OSGi R4. It appears the Log events spec has moved into its own chapter in later releases of the spec and all responsibility for the implementation should be in the felix.log component. Removing Event Admin’s Log Event adapter is necessary, otherwise in environment with both felix.log and felix.eventadmin the log events are duplicated.


wilx:
Rebased. Also, I misunderstood the spec initially. I restored some of the classes in EventAdmin and only moved the log related adapter.

PR #364, opened 2024-12-30
FELIX-6746 Lazy initialization for servlets extending JettyWebSocketServlet

paulrutter:
https://issues.apache.org/jira/projects/FELIX/issues/FELIX-6746

This PR adds lazy initialization for all servlets extending `JettyWebSocketServlet`.
This resolves the issue where [this](https://github.com/apache/felix-dev/blob/master/http/samples/whiteboard/src/main/java/org/apache/felix/http/samples/whiteboard/FelixJettyWebSocketServlet.java) was needed for earlier.

It also fixes the case where the legacy HTTP service is used in combination with the `ServletWrapper`.
The `ServletWrapper` instances are checked against wrapping a `JettyWebSocketServlet` as well.

Two integration tests added:
- using `JettyWebSocketServlet` via the whiteboard pattern. 
- using `JettyWebSocketServlet` via the HTTP service (by using `ServletWrapper`). This required passing along the websocket attributes to the servlet context, when `registerServlet` is called. This was already done for the whiteboard implementation earlier.

The whiteboard example has also been updated to reflect this.

@cziegeler what do you think of this approach?


paulrutter:
@enapps-enorman would you able to review this one? Since you handled the websocket functionality earlier as well.

http/base/src/main/java/org/apache/felix/http/base/internal/handler/ServletHandler.java
```diff
@@ -42,6 +44,9 @@ public abstract class ServletHandler implements Comparable<ServletHandler>
 
     private static final String JAVA_SERVLET_TEMP_DIR_PROP = "jakarta.servlet.content.tempdir";
```
cziegeler, line 46: How about moving the extra handling into a subclass and keep ServletHandler generic - and then make a decision when a new ServletHandler class is created whether this or the subclass should be used?

cziegeler:
The PR looks good to me in general, maybe the special handling could be moved into a subclass (see comment)?

http/base/src/main/java/org/apache/felix/http/base/internal/handler/ServletHandler.java
paulrutter: Good one, i will look into this later this week probably.

paulrutter: Can you take another look @cziegeler?

cziegeler (approved):
lgtm, thanks!

paulrutter:
I made one last change, moved a method and added a unit test.
Merged it now.

PR #367, opened 2025-01-10
[http] extract Multipart code into new ServletRequestMultipartWrapper

stbischof:
Mostly in the usage of felix-jetty-light there is the cases where you do not need the multipart support and can remove the dependencies of common-fileupload and common-io. To be able to do this we could not load a instantiate a class that imports common-fileupload classes. So we have to free the old class ServletRequestWrapper of this imports and add a ServletRequestMultipartWrapper that still have the multipart support.

### DIFF
--- a/http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
+++ b/http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
@@ -30,33 +30,18 @@ import static jakarta.servlet.RequestDispatcher.INCLUDE_REQUEST_URI;
 import static jakarta.servlet.RequestDispatcher.INCLUDE_SERVLET_PATH;
 import static org.apache.felix.http.base.internal.util.UriUtils.concat;
 
-import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.security.AccessControlContext;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import org.apache.commons.fileupload.FileItem;
-import org.apache.commons.fileupload.FileUpload;
-import org.apache.commons.fileupload.FileUploadBase;
-import org.apache.commons.fileupload.FileUploadException;
-import org.apache.commons.fileupload.RequestContext;
-import org.apache.commons.fileupload.disk.DiskFileItemFactory;
 import org.apache.felix.http.base.internal.context.ExtServletContext;
 import org.apache.felix.http.base.internal.handler.HttpSessionWrapper;
-import org.osgi.framework.Bundle;
 import org.osgi.service.servlet.context.ServletContextHelper;
 import org.osgi.service.useradmin.Authorization;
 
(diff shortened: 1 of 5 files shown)


paulrutter:
I haven't checked the whole PR yet, but i would expect `commons-fileupload` to be marked optional in the pom.xml of the base bundle.
Otherwise the bundle wouldn't even start if those classes are not on the classpath, right?

paulrutter:
I see, but wouldn't it make sense to mark the dependencies as optional as well, in https://github.com/apache/felix-dev/blob/f572bd8131df53e1fd39e9655afe6c7758541cc1/http/jetty/pom.xml#L504?

laeubi:
> but i would expect commons-fileupload to be marked optional in the pom.xml

Please note that "optional" means something totally different to maven, see https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

Optional is similar to "provided" but should not be included into runtime collections e.g. `slf4-api` will be `compile` and `slf4-simple` will be optional runtime because one might choose a different implementation of the API **but it wont work without any provided**!

Optional in OSGi mean it could work with or without this dependency, is really hard to get correct and almost never useful if not carefully crafted and tested. Especially in this case it might lead to hard to track runtime errors, so `ClassNotFound` Exceptions should be catched somewhere and output a meaningful message so the user understands how to resolve the issue.

paulrutter:
Thanks for clarifying @laeubi, that makes sense 👍

http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
```diff
@@ -411,143 +381,17 @@ public boolean isAsyncSupported()
...
-            this.parts.add(new PartImpl(item));
-        }
-    }
-
     @Override
     @SuppressWarnings({ "unchecked", "rawtypes" })
     public Collection<Part> getParts() throws IOException, ServletException {
-        return (Collection)checkMultipart();
-
+        return null;
```
cziegeler, line 387: According to the servlet spec, this method either throws a ServletException if the request is not a multipart request or always returns a collection. It never returns null. Throwing an exception seems to be the right thing

http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
```diff
@@ -411,143 +381,17 @@ public boolean isAsyncSupported()
...
     }
 
     @Override
     public Part getPart(String name) throws IOException, ServletException {
-        Collection<PartImpl> parts = this.checkMultipart();
-        for(final Part p : parts) {
-            if ( p.getName().equals(name) ) {
-                return p;
-            }
-        }
```
cziegeler, line 516: Similar to getParts this method should throw an exception

http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
stbischof: Thank you!
I now throw an ServletException.

http/base/src/main/java/org/apache/felix/http/base/internal/dispatch/ServletRequestWrapper.java
stbischof: I now throw an ServletException.

stbischof:
@cziegeler I now throw an ServletException on the 2 Messages. see my last commit.

cziegeler:
Thanks, looks good

stbischof:
also tested in my application. 
Now works proper without `file-upload` and `commons.io`

@paulrutter okay with a merge?

paulrutter:
@stbischof Yes 👍🏻 
The only thing i was wondering now is if the optional stated here (jetty12 bundle only) https://github.com/apache/felix-dev/blob/f572bd8131df53e1fd39e9655afe6c7758541cc1/http/jetty12/pom.xml#L738 is then correct.
From what i understood from @laeubi, that `optional` should not be there, right?

laeubi:
@paulrutter I think "optional" is sued semantically wrong here from maven point of view, but as the artifact is most likely never used as a maven dependency (but only as OSGi artifact it might not really matter, I would just remove all `<optional>` and let them be simple compile scope.

paulrutter:
@laeubi See https://github.com/apache/felix-dev/pull/368

PR #368, opened 2025-01-13
Remove Maven `optional` from Jetty bundle for `commons-io`

paulrutter:
Remove `optional` for `commons-io`, as the Jetty bundle is typically never used as maven dependency.
Optional means something different in maven than to OSGi. The optional OSGi instructions are left intact.

See discussion in https://github.com/apache/felix-dev/pull/367#issuecomment-2587205869

### DIFF
--- a/http/jetty/pom.xml
+++ b/http/jetty/pom.xml
@@ -465,11 +465,13 @@
             <groupId>org.eclipse.jetty.websocket</groupId>
             <artifactId>websocket-jakarta-server</artifactId>
             <version>${jetty.version}</version>
+            <optional>true</optional>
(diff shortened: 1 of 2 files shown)


stbischof:
this is not the only optional dep. in this pom

paulrutter:
Yes, my thought was also to remove all of them (see https://github.com/apache/felix-dev/pull/368/commits/5a29722e6e48a6de646b9b229d55cdc1533446f2). But that failed the ITs.
So there is at least some use for them right now.

paulrutter:
Closing in favor of https://github.com/apache/felix-dev/pull/505

PR #378, opened 2025-02-06
Upgrade Http Bridge Dependencies

StefanFranzWeiser:
The bridge imports on jakarta.servlet.descriptor avoids to deploy the bridge on Tomcat 11 correctly because Tomcat 11 provides v6.1. All other jakarta dependencies are open to exlude v7, see:
`
Import-Package: jakarta.servlet;version="[5.0,7)",jakarta.servlet.http
 ;version="[5.0,7)",org.osgi.service.useradmin;resolution:=optional;ve
 rsion="[1.1,2)",sun.misc;resolution:=optional,sun.nio.ch;resolution:=
 optional,jakarta.servlet.descriptor;version="[5.0,6)",java.io,java.la
 ng,java.lang.invoke,java.lang.ref,java.lang.reflect,java.math,java.ne
 t,java.nio,java.nio.channels,java.nio.charset,java.nio.file,java.nio.
`
Upgrading the bridge dependencies fixes the problem. Root cause is the reference on the old wrapper version 1.0.2.

### DIFF
--- a/http/bridge/pom.xml
+++ b/http/bridge/pom.xml
@@ -146,7 +146,7 @@
         <dependency>
             <groupId>jakarta.servlet</groupId>
             <artifactId>jakarta.servlet-api</artifactId>
-            <version>5.0.0</version>
+            <version>6.1.0</version>
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
@@ -162,13 +162,13 @@
         <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.service.http</artifactId>
-            <version>1.2.1</version>
+            <version>1.2.2</version>
            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>
             <artifactId>org.osgi.service.http.whiteboard</artifactId>
-            <version>1.1.0</version>
+            <version>1.1.1</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
@@ -179,17 +179,17 @@
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <version>2.15.0</version>
+            <version>2.18.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.http.base</artifactId>
-            <version>5.1.8</version>
+            <version>5.1.10</version>
         </dependency>
         <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.http.wrappers</artifactId>
-            <version>1.0.2</version>
+            <version>1.0.6</version>
 	    </dependency>
     </dependencies>
 </project>


http/bridge/pom.xml
```diff
@@ -179,17 +179,17 @@
...
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.http.base</artifactId>
-            <version>5.1.8</version>
+            <version>5.1.10</version>
         </dependency>
         <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.http.wrappers</artifactId>
-            <version>1.0.2</version>
+            <version>1.0.6</version>
```
paulrutter, line 192: 1.1.8 is latest for wrappers.

paulrutter:
Instead of moving the dependency to 6.1.0, wouldn't it suffice to add the version range in the import package?

```
jakarta.servlet.descriptor;version="[5.0,7)"
```

cziegeler:
The intention of the bridge is to run with both servlet 5 and servlet 6.  However,  servlet 5 needs wrappers 1.0.x and servlet 6 needs wrapper 1.1.x .
We can either create a second module (one for servlet 5 and one for servlet 6), or from now on we require at least servlet 6.
I think we should start with the latter and when users later request a bridge for servlet 5 create it on demand.

http/bridge/pom.xml
StefanFranzWeiser: Using 1.1.8 would cause breaking changes from API perspective (build failure). Therefore, I sticked on 1.0.6, the latest compatible one.

StefanFranzWeiser:
> Instead of moving the dependency to 6.1.0, wouldn't it suffice to add the version range in the import package?
> 
> ```
> jakarta.servlet.descriptor;version="[5.0,7)"
> ```

The version range is not defined by the bridge, instead inherited from the wrappers.

StefanFranzWeiser:
> The intention of the bridge is to run with both servlet 5 and servlet 6. However, servlet 5 needs wrappers 1.0.x and servlet 6 needs wrapper 1.1.x . We can either create a second module (one for servlet 5 and one for servlet 6), or from now on we require at least servlet 6. I think we should start with the latter and when users later request a bridge for servlet 5 create it on demand.

On my side wrappers 1.0.6 is enough and repairs the version range for my purpose. At the moment, the bridge cannot be used for servlet v6 because jakarta.servlet.descriptor limits it to v5, while jakarta.servlet and jakarta.servlet.http are open to v5 and v6.

Tomcat 9 is still on javax.
Tomcat 10 uses jakarta servlet 6.0.
Tomcat 11 uses jakarta servlet 6.1.

cziegeler:
While it might work, still wrappers 1.0.x is for servlet 5 and 1.1.x is for servlet 6.
If we just support servlet 6 with the bridge then we need of course to change the version range for all servlet packages to be [6,7),

StefanFranzWeiser:
If wrappers 1.0.x are for servlet 5 and 1.1.x is for servlet 6, does the bridge then only support servlet 5 at the moment, not 6? This would not be cool :-(

cziegeler:
As long as you don't use some of the new methods from servlet 6, the bridge will work with servlet 5. As soon as you try to use those methods it will fail *if* the wrappers are needed during a request. Which means if there is some code relying on javax.servlet. The combination of code using javax.servlet and at the same time using servlet 6 methods is very unlikely to happen, therefore you probably do not notice any problems. But still its possible to fail.

StefanFranzWeiser:
Just identified that http.proxy is also limited to jakarta servlet 5. When upgrading to 6 only the ServletContextWrapper does not compile which is adapted fast.

Summing up, looks like felix http does not support jakarta servlet 6 totally at the moment. 😞 
Therefore, will adapt it for my purpose and base on my fork instead (and keep waiting for new release 😄).

StefanFranzWeiser:
The updated proxy and bridge works from dependency management. By the way, I saw that the ProxyListener was removed (also mentioned in Tutorials). Was this intended? I changed it to ProxyServletContextListener, hoping everything works as intended.

cziegeler:
Thanks for the feedback. Yes, ProxyListener has been replaced by ProxyServletContextListener

StefanFranzWeiser:
Solved with new v6.0.0, thx to @cziegeler

PR #379, opened 2025-02-08
FELIX-6754 Update jetty to 12.0.17

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6754


paulrutter:
Added logic for handling new `JETTY_11` uri compliance mode, see https://github.com/jetty/jetty.project/pull/12827

PR #383, opened 2025-02-20
Do not print an error when a wildcard is used in component header

laeubi:
Currently SCR prints an error if one uses a wildcard in the header but nothing matches. But because fragemnts are not allowed to provide an own header, only the host can enable this.

So this now adds an additional check before logging an error that prevents an error from being printed if a wildcard is provided.

See

- https://github.com/osgi/osgi/issues/789
- https://github.com/eclipse-platform/.github/issues/241
- https://github.com/eclipse-lemminx/lemminx/pull/1732

### DIFF
--- a/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
+++ b/scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
@@ -270,11 +270,29 @@ public class BundleComponentActivator implements ComponentActivator
                 URL[] descriptorURLs = findDescriptors(m_bundle, descriptorLocation);
                 if (descriptorURLs.length == 0)
                 {
-                    // 112.4.1 If an XML document specified by the header cannot be located in the bundle and its attached
-                    // fragments, SCR must log an error message with the Log Service, if present, and continue.
-                    logger.log(Level.ERROR,
-                        "Component descriptor entry ''{0}'' not found", null,
-                        descriptorLocation);
+                    if (descriptorLocation.contains("*")) {
+                        // 112.4.1 The last component of each path in the Service-Component header may
+                        // use wildcards so that Bundle.findEntries can be used to locate the XML
+                        // document within the bundle and its fragments. For example:
+                        //
+                        // Service-Component: OSGI-INF/*.xml
+                        //
+                        // A Service-Component manifest header specified in a fragment is ignored by
+                        // SCR. However, XML documents referenced by a bundle's Service-Component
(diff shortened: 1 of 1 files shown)


scr/src/main/java/org/apache/felix/scr/impl/BundleComponentActivator.java
```diff
@@ -270,6 +270,21 @@ protected void initialize(List<ComponentMetadata> cachedComponentMetadata)
...
+                        //
+                        // Service-Component: OSGI-INF/*.xml
+                        //
+                        // A Service-Component manifest header specified in a fragment is ignored by
+                        // SCR. However, XML documents referenced by a bundle's Service-Component
+                        // manifest header may be contained in attached fragments.
+
+                        // in case of such wildcard, finding nothing does not mean an error (because it
+                        // *might* be found in a fragment that is attached later.
+                        continue;
```
tjwatson, line 286: Perhaps we should log as TRACE here?  By default that should get ignored since the default log level is WARN.

laeubi: Done!

tjwatson (approved):

stbischof:
What i the benefit?

We log an error before. Now we also trace the same information additional.

Isnt that just Duplikat?

laeubi:
Please notice this line in the PR:

https://github.com/apache/felix-dev/pull/383/files#diff-1cdeca2eb3dfd6054d5c998e7acdafef598a9a30bab68ebc4912aac55d8f74ffR289

so before it always printed an error, now it only prints error if not a wildcard (but a trace message instead).

dpolivaev:
Is there any movement on it?
Current state produces misleading error messages in eclipse log, and nobody cares there too because they believe it is apache felix issue.

https://github.com/eclipse-lemminx/lemminx/pull/1732
https://github.com/eclipse-lemminx/lemminx/issues/1731

This is how you loose trust.

laeubi:
> This is how you loose trust.

Trust in what? This is all open source so "nobody" is just "everybody" ...

tjwatson (approved):

tjwatson:
> damn tj was faster with merge ;-)

Now someone has to sign up to do a release; otherwise we loose all trust ;-)

andrew-tram:
It looks like this was fixed as part of org.apache.felix.scr 2.2.13? When will that be able? I dont see it floating around Orbit or other repos just yet.

PR #392, opened 2025-03-13
FELIX-6758 Update to Jetty 12.0.18 / 11.0.25

paulrutter:
- Upgrade to 12.0.18 and 11.0.25


joakime:
Both of those versions of Jetty will be available on Maven Central today.

paulrutter:
Thanks for sharing @joakime 👍

joakime:
I can confirm that both versions are now present on Maven Central.
Good luck!

PR #399, opened 2025-04-02
FELIX-6761 : NullPointerException when registering a new servlet

cziegeler:


cziegeler:
@paulrutter I merged the two servlet handler classes into a single one to avoid instantiating the servlet object twice (and before init is called)

paulrutter:
I will check it out this morning 👍

paulrutter:
@cziegeler I checked out the changes, but i see the WS related unit test now fails.
To your points:

* it does not check that getting the servlet might result in null -> this is clear
* it gets the servlets twice, at least in the case of normal whiteboard handling -> clear, but the behavior is not the same as before

The unit tests fail now:

```
[INFO]
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   JettySpecificWebsocketIT.testWebSocketServletWhiteboard ┬╗ Upgrade Failed to upgrade to websocket: Unexpected HTTP Response Status Code: 500 Server Error
[INFO]
[ERROR] Tests run: 10, Failures: 0, Errors: 1, Skipped: 0
[INFO]
```

paulrutter:
@cziegeler I fixed it in https://github.com/apache/felix-dev/pull/399/commits/34ec76fcc99a3cad2ed6cdece5a57c74e128a0c5, the unit tests pass now 👍🏻

cziegeler:
@paulrutter Thanks! yes, I forgot to run the tests :)

paulrutter:
LGTM 👍🏻

paulrutter:
@cziegeler ps. in which situation did the NPE actually occur? I haven't run into this myself.

PR #402, opened 2025-04-04
[FELIX-6765] - Add a CI-job for Felix Framework

stbischof:
- ci job
- update mockito-core to mockito to latest version to run on java 17 and not only 8
- update felix-parent version to avoid build issue with `org.codehaus.mojo:ianal-maven-plugin:1.0-alpha-1`

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -10,6 +10,7 @@ on:
       - 'tools/osgicheck-maven-plugin/**'
       - 'log/**'
       - 'webconsole/**'
+      - 'framework/**'
   pull_request:
     branches: [ "master" ]
     paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/maven-ci.yml
```diff
@@ -27,10 +29,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Set up JDK 17
+    - name: Set up JDK 8
       uses: actions/setup-java@v4
       with:
-        java-version: '17'
+        java-version: '8'
```
paulrutter, line 35: This will break HTTP and other projects requiring at least 17.

stbischof: you are right,
will try to fix all test that need jdk 8 in framework

https://github.com/apache/felix-dev/actions/runs/14275725289/job/40018259906#step:12:1151

paulrutter (approved):
LGTM!

PR #403, opened 2025-04-05
FELIX-6766 Improvement Tests of framework

stbischof:
- junit 4-> junit.jupiter
- update to latest Mockito version
- remove hamcrest
- remove easymock
- use assertj
- use current code conventions

PR #405, opened 2025-04-05
[framework] add tck module - run the OSGi-TCK osgi.test.cases.framework 

stbischof:
first commit could be ignored. ist PR #403

relevant is the 2. commit that showns running  OSGi-TCK osgi.test.cases.framework 

in actions you can see that framework is tck complient

https://issues.apache.org/jira/browse/FELIX-6766

### DIFF
--- a/.github/workflows/maven-ci.yml
+++ b/.github/workflows/maven-ci.yml
@@ -27,12 +27,16 @@ permissions: {}
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [17]
     steps:
     - uses: actions/checkout@v4
-    - name: Set up JDK 17
+    - name: Set up JDK ${{ matrix.java }}
       uses: actions/setup-java@v4
       with:
-        java-version: '17'
+        java-version: ${{ matrix.java }}
         distribution: 'temurin'
         cache: maven
     - name: Set up Maven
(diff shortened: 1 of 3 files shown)


framework.tck/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+    <dependency>
+      <groupId>org.apache.felix</groupId>
+      <artifactId>org.apache.felix.framework</artifactId>
+      <version>7.1.0-SNAPSHOT</version>
+    </dependency>
+    <!-- tck run -->
+    <dependency>
+      <groupId>biz.aQute.bnd</groupId>
+      <artifactId>biz.aQute.junit</artifactId>
+      <version>6.4.1</version>
```
laeubi, line 156: Should this not 7.0?

laeubi:
@stbischof for Equinox I let it generate some nice badges you can see the results here:

https://github.com/eclipse-equinox/equinox?tab=readme-ov-file#implemented-services-an-compliance

I also noticed that here it is mentioned felix implement R7:

https://github.com/apache/felix-dev?tab=readme-ov-file#felix-framework

framework.tck/pom.xml
stbischof: or 7.1.0 thx

.github/workflows/maven-ci.yml
```diff
@@ -27,12 +27,16 @@ permissions: {}
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: true
```
paulrutter, line 30: Will this ignore any build or test failures? That would not be ideal for CI.

stbischof: Thank you, you found the leftover of a test, will remove.

stbischof: Will Switch to fail-fast: false that we can run in matrix against multiple versions and nonjob will break when an other is braking

PR #409, opened 2025-04-13
Gogo ci

stbischof:


.github/workflows/maven-ci.yml
```diff
@@ -87,6 +90,9 @@ jobs:
     - name: OSGi-TCK Framework
       if: steps.changes.outputs.framework == 'true'
       run: mvn -B -V -Dstyle.color=always --file framework.tck/pom.xml clean verify
+    - name: Felix Gogo Shell
+      if: steps.changes.outputs.gogo == 'true'
+      run: mvn -B -V -Dstyle.color=always --file gogo/pom.xml clean install verify
```
laeubi, line 95:
```suggestion
      run: mvn -B -V -Dstyle.color=always --file gogo/pom.xml clean verify
```

`verify` includes `install`...

gogo/itest-shell/generated/launch9110451383565747987.properties
laeubi, line 1: Should `generated` really be checked in?

stbischof: Niope. Will set a gitignore

.github/workflows/maven-ci.yml
stbischof: Thx for finding. Will update

laeubi:
@paulrutter can we merge this?

.github/workflows/maven-ci.yml
dmatej: It is the opposite, install includes verify. See https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html
I just noticed this in email ;-)

paulrutter:
> @paulrutter can we merge this?

Once review comments are processed, @stbischof can merge it 👍🏻

.github/workflows/maven-ci.yml
stbischof: You are wright. But. Why should i install

laeubi: Yes sorry for confusion, I think verify is enough (as mentioned in the suggestion) as all other tasks are also use verify to execute the tests.

stbischof:
removed install in mvn command and the generated folder.

PR #412, opened 2025-04-29
Fix interruption handling in nested shells

gnodet:
# Fix interruption handling in nested shells

This PR complements PR #411 to fully fix the issue reported in jline/jline3#1143 where interruption exceptions (Ctrl+C) were not working for child sessions.

## Problem
When a user starts a shell session, then runs the 'sh' command to create a nested shell, and then runs a command like 'ttop' in that nested shell, pressing Ctrl+C doesn't properly interrupt the command.

## Solution
This PR modifies the Shell.sh method to properly handle interruption signals in nested shells by clearing and restoring the current pipe. The fix follows the same pattern implemented in the Posix.runShell method, ensuring consistent behavior across different shell creation methods.

PR #411 made the Pipe.setCurrentPipe method public, which is a prerequisite for this fix. Together, these changes ensure that interruption signals are properly propagated to commands running in nested shells.

### DIFF
--- a/gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
+++ b/gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
@@ -71,6 +71,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.felix.gogo.runtime.Pipe;
 import org.apache.felix.service.command.Process;
 import org.apache.felix.gogo.jline.Shell.Context;
 import org.apache.felix.service.command.CommandProcessor;
@@ -811,6 +812,10 @@ public class Posix {
     private void runShell(CommandSession session, Terminal terminal) {
         InputStream in = terminal.input();
         OutputStream out = terminal.output();
+        
+        // Save the current pipe and clear it before creating a child shell
+        Pipe currentPipe = Pipe.setCurrentPipe(null);
+
         CommandSession newSession = processor.createSession(in, out, out);
         newSession.put(Shell.VAR_TERMINAL, terminal);
         newSession.put(".tmux", session.get(".tmux"));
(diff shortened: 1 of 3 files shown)


gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
```diff
@@ -808,9 +809,26 @@ private void startShell(CommandSession session, Terminal terminal) {
         new Thread(() -> runShell(session, terminal), terminal.getName() + " shell").start();
     }
 
+    /**
```
stbischof, line 812: Do we really need this doc on a private messge whn al is written in commit message

stbischof:
could you please rebase on the merged commit

commit-message.txt
paulrutter, line 1: This can just be part of the commit message, doesn't have to be a git artifact

gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
paulrutter: Agreed, see my previous comment

gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
```diff
@@ -822,11 +840,26 @@ public void exit() throws Exception {
                 terminal.close();
             }
         };
+
+        // Register a signal handler for INT signal to properly propagate interruption
+        Terminal.SignalHandler prevIntHandler = terminal.handle(Terminal.Signal.INT, signal -> {
```
paulrutter, line 845: is `Terminal.Signal.INT` working cross-platform? E.g. also on Windows? I know Windows can sometimes be different in handling signals.

gogo/runtime/src/main/java/org/apache/felix/gogo/runtime/Pipe.java
```diff
@@ -107,7 +107,12 @@ public static Pipe getCurrentPipe() {
         return CURRENT.get();
     }
 
-    private static Pipe setCurrentPipe(Pipe pipe) {
+    /**
+     * Set the current pipe for the current thread.
+     * @param pipe the pipe to set as current, or null to clear
+     * @return the previous pipe
+     */
+    public static Pipe setCurrentPipe(Pipe pipe) {
```
paulrutter, line 115: rebase the PR on master to remove this change.

paulrutter (changes requested):

commit-message.txt
gnodet: This wasn't intended to be part of the commit ;-)

gogo/jline/src/main/java/org/apache/felix/gogo/jline/Posix.java
gnodet: It definitely is.

paulrutter (approved):

PR #413, opened 2025-04-30
FELIX-6773 Fixing an issue which is causing unnecessary downloads

RefuX:
FELIX-6773 Fixing an issue which is causing unnecessary downloads in ResolverImpl due to resources that are equals being reported as unequal.

https://issues.apache.org/jira/browse/FELIX-6773

### DIFF
--- a/bundlerepository/src/main/java/org/apache/felix/bundlerepository/impl/LazyLocalResourceImpl.java
+++ b/bundlerepository/src/main/java/org/apache/felix/bundlerepository/impl/LazyLocalResourceImpl.java
@@ -115,4 +115,8 @@ public class LazyLocalResourceImpl implements LocalResource
     public boolean equals(Object o) {
         return getResource().equals(o);
     }
+
+    public int hashCode() {
+        return getResource().hashCode();
+    }
 }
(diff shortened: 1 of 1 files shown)


stbischof:
Could you please stash this commits together

bundlerepository/src/main/java/org/apache/felix/bundlerepository/impl/LazyLocalResourceImpl.java
```diff
@@ -111,4 +111,12 @@ public Capability[] getCapabilities() {
     public Requirement[] getRequirements() {
         return getResource().getRequirements();
     }
+
```
stbischof, line 114: please use the Override Annotation

bundlerepository/src/main/java/org/apache/felix/bundlerepository/impl/LazyLocalResourceImpl.java
```diff
@@ -111,4 +111,12 @@ public Capability[] getCapabilities() {
...
+    public boolean equals(Object o) {
+        return getResource().equals(o);
+    }
+
```
stbischof, line 118: please use the Override Annotation

stbischof (changes requested):
I do no think, that we need 3 commits for this lines. Could you also add a Test

stbischof:
@laeubi comparing Resources, do you know more?

PR #414, opened 2025-05-03
FELIX-6773 Fixing an issue which is causing unnecessary downloads

RefuX:
FELIX-6773 Fixing an issue which is causing unnecessary downloads in ResolverImpl due to resources that are equals being reported as unequal.

https://issues.apache.org/jira/browse/FELIX-6773


stbischof:
incrementation of this:
https://github.com/apache/felix-dev/pull/413

stbischof:
Could you please rebase that we just habe the one commit.

PR #416, opened 2025-05-07
FELIX-6774 `org.apache.felix.http.jetty.maxFormSize` not enforced: Fix

paulrutter:
On top of https://github.com/apache/felix-dev/pull/415 which demonstrates the issue at hand.

- Introducing two new config properties, `org.apache.felix.http.jetty.requestSizeLimit` and `org.apache.felix.http.jetty.responseSizeLimit`
- Deprecate `org.apache.felix.http.jetty.maxFormSize` as it doesn't seem to enforce the expected limit
- If either one is configured, the `SizeLimitHandler` will be configured, see https://jetty.org/docs/jetty/12/programming-guide/server/http.html#handler-use-size-limit
- -1 is the default for both, which means unlimited.
- Do we still need to set `setMaxFormContentSize` now?
- Updated README.md


paulrutter:
Rejecting this PR, as it doesn't belong to FELIX-6774.

PR #418, opened 2025-05-07
FELIX-6776 introduce way to restrict request and response size

paulrutter:
See https://issues.apache.org/jira/browse/FELIX-6776 and https://issues.apache.org/jira/browse/FELIX-6774 which led to creating this issue.


cziegeler (approved):
Lgtm, thanks

PR #425, opened 2025-05-17
FELIX-6781 ResolverImpl should start all deployed resources (if START flag used)

RefuX:
FELIX-6781 ResolverImpl should start all deployed resources (if START flag used)

https://issues.apache.org/jira/browse/FELIX-6781


stbischof:
@bosschaert @cziegeler 

Any thoughts on this and the other pr that i related to bundlerepo?

rickhall:
I think it seems reasonable. We treat local bundles as being part of a local repository, so you could argue it is consistent to treat them the same way (with regards to the START flag) as bundles from a remote repo.

PR #426, opened 2025-05-19
FELIX-6782 allow adding custom headers to jetty error pages

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6782


paulrutter:
@cziegeler Any feedback on this PR? 
I need this to add `Strict-Transport-Security` to error pages, for security reasons.

http/jetty12/src/main/java/org/apache/felix/http/jetty/internal/ConfigMetaTypeProvider.java
```diff
@@ -220,6 +220,12 @@ public ObjectClassDefinition getObjectClassDefinition( String id, String locale
                 204800,
                 bundle.getBundleContext().getProperty(JettyConfig.FELIX_JETTY_RESPONSE_SIZE_LIMIT)));
 
+        adList.add(new AttributeDefinitionImpl(JettyConfig.FELIX_JETTY_ERROR_PAGE_CUSTOM_HEADERS,
+                "Custom headers to add to error pages",
+                "Felix specific property to configure the custom headers to add to all error pages served by Jetty. Separate key-value pairs with ##.",
+                204800,
```
paulrutter, line 226: I need to adjust this value to the right defaults.

http/jetty12/src/main/java/org/apache/felix/http/jetty/internal/JettyConfig.java
```diff
@@ -509,6 +512,10 @@ public int getResponseSizeLimit()
         return getIntProperty(FELIX_JETTY_RESPONSE_SIZE_LIMIT, -1);
     }
 
+    public String getFelixJettyErrorPageCustomHeaders() {
+        return getProperty(FELIX_JETTY_ERROR_PAGE_CUSTOM_HEADERS, null);
```
paulrutter, line 516: Not using `getStringArray` as splitting on "," is not safe for header values; these might contain "," in the value as well. For example:

`Expires=Thu, 01 Dec 1994 16:00:00 GMT`

cziegeler (approved):
Makes sense and looks good to me, thanks

PR #427, opened 2025-05-26
FELIX-6784 Reduce number of thread (re)creations of the Configurator Worker Thread

glimmerveen:
Changed the WorkerQueue to be based on the Java's standard ScheduledThreadPoolExecutor. This already implements keeping track of tasks, and allows for the configuring a keep alive time.

### DIFF
--- a/configurator/src/main/java/org/apache/felix/configurator/impl/WorkerQueue.java
+++ b/configurator/src/main/java/org/apache/felix/configurator/impl/WorkerQueue.java
@@ -18,67 +18,32 @@
  */
 package org.apache.felix.configurator.impl;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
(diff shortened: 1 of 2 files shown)


configurator/src/main/java/org/apache/felix/configurator/impl/WorkerQueue.java
```diff
@@ -18,67 +18,32 @@
...
+            @Override
+            public Thread newThread(Runnable runnable) {
+                Thread thread = new Thread(runnable);
+                thread.setDaemon(true);
+                thread.setName("Apache Felix Configurator Worker Thread");
+                return thread;
             }
-        }
+        });
+        setKeepAliveTime(5, TimeUnit.SECONDS);
```
glimmerveen, line 39: This could be made configurable, but my initial proposal is to have a static, relatively short, keep alive time fixed in the code. Its purpose is mainly to let the underlying thread terminate at some point if no tasks are queued.

stbischof:
Hi, did you run this against the tck?

Can you rebase your pr and push again? Then some actions should start the test.

PR #430, opened 2025-07-03
FELIX-6791 Only a single format can be allowed when using Pax Web

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6791

- Add unique servlet name to each registered servlet via the whiteboard pattern


cziegeler:
CLosing as wont fix

PR #433, opened 2025-07-14
FELIX-6807 - Java 25 LTS support

paulrutter:
Try-out building framework and HTTP subprojects against java 25 to see what will break

https://issues.apache.org/jira/browse/FELIX-6807


paulrutter:
> can you also do a change in scr to trigger this test

Will do tomorrow 👍

stbischof:
common solution between Equinox and Felix .... And springboot would be really best way to solve.

stbischof:
I disagree with your comment in Security manager.
There are users that use lower Versions if java and need Security Management.

In all cases where Security Managemer is uses we check Existense before we use that. So no execution in Versions where Security Manager is remived.

laeubi:
@mattrpav Just wanted to note that JDT has now new support for multi-release jars that maps nicely to what we have in maven.

Multi-Release Jars are a perfect fit for such kind of support such JDK dependent changes, then one only need a `SecurityManagerFacade` that is a multi-release type where depending on the JVM version either use security manager or make it a no-op on JDK25+

mattrpav:
@tjwatson I mispoke-- JDK 25 _disables_ the SecurityManager by default and custom SecurityManages cannot be installed.

@laeubi MRJ is a good idea, will check that out.

mattrpav:
> I have a distaste for MRJ

Anything specific? I'm looking at using a MRJ for activemq-client to use that for the Virtual Thread classes that need JDK 21+.

tjwatson:
> But this will only work if the API is not gone (or non existent) if you not choose reflection ... what is a nightmare on its own of course.

This is all theoretical since `SecurityManager` is still there.  I would never recommend going to MRJs if the class is still there.

I also would never recommend using a compiler that doesn't understand the release you are targeting where the class still exists that your code wants to use on old Java versions.  Even if/when they really remove the class the `--release` option must continue to know about the removed class for the java version you are targeting and must allow your code to compile against it.  If no compiler exists that can do the `--release` option to the java version you are targeting then it is really finally time you move onto a new Java version.

Since Oracle just [extended their Java 8 support](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) way out to 2030 (and other vendors will follow) I don't see that happening any earlier for the new Java versions being released through 2030.  I suspect they will continue to support `--release 8` for a long time yet.

> The source jar?!? Do you mean the final build jar or the source code in the project?

I mean the source jar, typically published to maven central.  As in [org.eclipse.osgi-3.23.200-sources.jar](https://repo1.maven.org/maven2/org/eclipse/platform/org.eclipse.osgi/3.23.200/org.eclipse.osgi-3.23.200-sources.jar)

Even if that all works out and all IDEs handle it in a useful way. I often am handed an error log and asked to analyze some stack trace with methods and line numbers.  At that point it is nice to just know the version I am dealing with to find the code for the class/method/line-number in the stack trace without having to worry about which "version" of the class within the MRJ got loaded.

I am doubtful anyone can convince me that MRJs are a perfect solution for me.

tjwatson:
> Anything specific? I'm looking at using a MRJ for activemq-client to use that for the Virtual Thread classes that need JDK 21+.

Even for that I would use a separate project that holds the Java 21+ code that uses Virftual Threads and then pull the classes into the project that still targets the old Java release.  Then have code that knows what class to load based on Java versions.  That is what we do for the Virtual threads.  We then created an `org.osgi.service.condition.Condition` ID for the Java version so that we can enable the correct service component based on the Java version.

For example, this component only gets enabled for Java versions < 21:
https://github.com/OpenLiberty/open-liberty/blob/f8638a00c65c9dec748d21253571b2d0acd9fcab/dev/com.ibm.ws.threading/src/com/ibm/ws/threading/internal/VirtualThreadDisallowed.java#L34

And this one gets enabled for Java versions >= 21

https://github.com/OpenLiberty/open-liberty/blob/f8638a00c65c9dec748d21253571b2d0acd9fcab/dev/io.openliberty.threading.virtual.internal/src/io/openliberty/threading/virtual/internal/VirtualThreadOperations.java#L40

The rest of the system gets injected with the right component implementation without having to worry about what Java version is running.

paulrutter:
> > Anything specific? I'm looking at using a MRJ for activemq-client to use that for the Virtual Thread classes that need JDK 21+.
> 
> Even for that I would use a separate project that holds the Java 21+ code that uses Virftual Threads and then pull the classes into the project that still targets the old Java release. Then have code that knows what class to load based on Java versions. That is what we do for the Virtual threads. We then created an `org.osgi.service.condition.Condition` ID for the Java version so that we can enable the correct service component based on the Java version.
> 
> For example, this component only gets enabled for Java versions < 21: https://github.com/OpenLiberty/open-liberty/blob/f8638a00c65c9dec748d21253571b2d0acd9fcab/dev/com.ibm.ws.threading/src/com/ibm/ws/threading/internal/VirtualThreadDisallowed.java#L34
> 
> And this one gets enabled for Java versions >= 21
> 
> https://github.com/OpenLiberty/open-liberty/blob/f8638a00c65c9dec748d21253571b2d0acd9fcab/dev/io.openliberty.threading.virtual.internal/src/io/openliberty/threading/virtual/internal/VirtualThreadOperations.java#L40
> 
> The rest of the system gets injected with the right component implementation without having to worry about what Java version is running.

That's an elegant solution to the problem indeed. In Felix, we have resorted to reflection to determine java 21 specific features, see for example https://github.com/apache/felix-dev/blob/0c74c4693d49211beff294a2cfdf5bf42c8c0e14/http/jetty12/src/main/java/org/apache/felix/http/jetty/internal/JettyService.java#L287. But this is pretty isolated.

To get back to the problem of the SecurityManager in Felix, i feel we've only scratched the surface and are not aligned on the approach to take. The options as i see them:
1. use reflection or a similar approach outlined above to go about distinguishing java features and thus remain to have one main branch
2. Maintain two branches, depending on java version support or remain to have one and increase minimum java version to 25 (probably undesired) in next release
3. Investigate MJR feasibility
4. Other ideas? 

I don't have a strong preference yet, but i do think we should come up with a plan forward, while also respecting the [OSGi spec](https://osgi.github.io/osgi/core/framework.security.html) on the security aspects.

laeubi:
> I am doubtful anyone can convince me that MRJs are a perfect solution for me.

I don't want to convince anyone, just mentioning that its maybe worth to consider and most of what you describe more feels like limitations of the tooling that could probably be better enhanced there.

Compared to that your solution seem to add a lot of complexity both on the build time as well as dev time just to emulate a feature the JVM can already offer to you (you then need to make sure to have at least a MR OSGi-Manifest), but again I don't want to convince anyone and there are obviously multiple ways.

The problem of multi-release source jars is discussed here so you maybe want to leave a not if you thing its not feasible (even though you not plan to use it) as likely others might still adopt:

- https://github.com/apache/maven-source-plugin/issues/215

ben-manes:
I am trying to get a basic pax-exam unit test to work on jdk 25 and it seems to fail due to Felix not being compatible, if I understand correctly. https://github.com/ops4j/org.ops4j.pax.exam/issues/1131

```console
org.osgi.framework.BundleException: Unable to resolve com.github.ben-manes.caffeine [15](R 15.0): missing requirement [com.github.ben-manes.caffeine [15](R 15.0)] osgi.ee; (osgi.ee=UNKNOWN) Unresolved requirements: [[com.github.ben-manes.caffeine [15](R 15.0)] osgi.ee; (osgi.ee=UNKNOWN)]
	at org.apache.felix.framework.Felix.resolveBundleRevision(Felix.java:4398)
	at org.apache.felix.framework.Felix.startBundle(Felix.java:2308)
	at org.apache.felix.framework.Felix.setActiveStartLevel(Felix.java:1566)
	at org.apache.felix.framework.FrameworkStartLevelImpl.run(FrameworkStartLevelImpl.java:297)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

I tried passing in `org.osgi.framework.executionenvironment` as `JavaSE-11` but no luck.

JnRouvignac:
Hello!

IIUC, this PR is blocked waiting on `bnd` to support JDK 25?
The 7.2.0 release of `bnd` says it supports it: https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0

Does it unblock this PR and support of JDK 25 in felix?

paulrutter:
> Hello!
> 
> IIUC, this PR is blocked waiting on `bnd` to support JDK 25? The 7.2.0 release of `bnd` says it supports it: https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0
> 
> Does it unblock this PR and support of JDK 25 in felix?

No, it doesn't. The work for supporting JDK 25 haven't really started yet, see https://github.com/apache/felix-dev/pull/433#issuecomment-3357792022. The biggest hurdle i see is resolving the removed `SecurityManager` usage from the codebase, while retaining compatibility with older JDK's.

JnRouvignac:
> > Hello!
> > IIUC, this PR is blocked waiting on `bnd` to support JDK 25? The 7.2.0 release of `bnd` says it supports it: https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0
> > Does it unblock this PR and support of JDK 25 in felix?
> 
> No, it doesn't. The work for supporting JDK 25 haven't really started yet, see [#433 (comment)](https://github.com/apache/felix-dev/pull/433#issuecomment-3357792022). The biggest hurdle i see is resolving the removed `SecurityManager` usage from the codebase, while retaining compatibility with older JDK's.

OK, thanks a lot for clarifying this.
I was a bit lost on the status of this issue with the many comments.

JnRouvignac:
@paulrutter I am interested in seeing this issue resolved.
Is there anything I can do to help unblock the situation? (Loose promise, if I can get a chance)

I saw you're considering several possible solutions for the `SecurityManager`:
* use reflection
* maintain two branches
* multi-release jar
* other?

What would help in this evaluation?

JnRouvignac:
@paulrutter I am interested in seeing this issue resolved.
Is there anything I can do to help unblock the situation? (Loose promise, if I can get a chance)

I saw you're considering several possible solutions for the `SecurityManager`:

* use reflection
* maintain two branches
* multi-release jar
* other?

What would help with this evaluation?

mattrpav:
Many other projects settled on using a MR jar and a SubjectShim class that wraps the differences in the JAAS api for resolving the Subject

paulrutter:
Given [this comment](https://github.com/apache/felix-dev/pull/433#issuecomment-3357572321) from @tjwatson, i doubt MR's are the way forward.

* Reflection could work, but bloat the codebase. Not ideal, but could work.
* two branches seems like a viable solution; the main branch could contain the "latest and greatest" support (JDK 25, sans `SecurityManager`), while maintaining a branch for older JDK's. That the approach we're taking for the HTTP subproject for EE10 and EE11 as well. This would be my preferred approach.
** maintaining two branches carries a maintenance burden, but at least the core framework doesn't change often, so that risk we can probably afford to have.

I would like to get some more opinions (and hopefully consensus) on this before we start work in either direction.

mbien:
> * two branches seems like a viable solution

fwiw: JDK recommends the two branch model too https://openjdk.org/jeps/14

paulrutter:
## Status update: SecurityManager removed, JDK 25 build blockers fixed, TCK green, framework at 8.0.0

Picking up the two-branch approach discussed in [#433 (comment)](https://github.com/apache/felix-dev/pull/433#issuecomment-4543766603) and confirmed by @mbien's [JEP 14 reference](https://github.com/apache/felix-dev/pull/433#issuecomment-5407380786): **this branch drops the Security Manager entirely and fixes what actually breaks on JDK 25.** The branch has also been synced with master.

Everything below was verified locally on **JDK 25.0.2**, each time against a **JDK 21 control run on the same tree**, so that pre-existing and platform-specific failures were not misattributed to JDK 25.

### What actually blocked JDK 25

Worth recording, because it was much narrower than expected:

1. `framework/pom.xml` passed `-Djava.security.manager=allow`. On JDK 25 that is a fatal **VM startup** error (JEP 486), so **zero tests ran** — this one flag was masking everything else.
2. Once removed, the *only* JDK-25-attributable failures were the two `URLHandlersTest` SecurityManager tests.

Probing the API directly on JDK 25: only `System.setSecurityManager` actually throws. `AccessController.doPrivileged`, `getContext`, `Policy.getPolicy`, `ProtectionDomain` — and even subclassing `SecurityManager` — all still work. So the framework's sole *production* blocker was `Felix.init()`; the rest is dead code that could never execute again.

### SecurityManager removal (−8,000 lines)

Enforcement is gone; **the OSGi API types are kept**, so bundles referencing `AdminPermission`, `ServicePermission`, `PackagePermission`, `Bundle.hasPermission`, `ProtectionDomain` etc. still compile and resolve.

- `Felix.init()` no longer installs a Security Manager. Setting `org.osgi.framework.security` now **fails fast** with a `SecurityException` rather than silently launching without the security the launcher asked for.
- `SecureAction`: **2113 → 656 lines**; all 65 `doPrivileged` wrappers collapsed to their direct calls, the `Actions` dispatch class deleted.
- `SecurityManagerEx` no longer extends `SecurityManager` — it now uses `StackWalker`, the supported replacement for `getClassContext()`, which yields frames in the same order. No `extends SecurityManager` remains anywhere.
- Enforcement removed from `BundleImpl`, `BundleContextImpl`, `BundleWiringImpl`, `EventDispatcher`, `WovenClassImpl`, `StatefulResolver`, `ExtensionManager`, `ServiceRegistrationImpl`, `FrameworkStartLevelImpl`, `FrameworkWiringImpl`, `URLHandlersBundleStreamHandler`, `BundleProtectionDomain`, plus `scr`, `webconsole` and `gogo`. Every removed branch was already unreachable without a Security Manager, so behaviour is unchanged.
- **`framework.security` removed** — it existed only to implement the SM-based `SecurityProvider`.

**`felix.java.version` for `framework` is now 9** — the *lowest* level it still builds at. To be explicit: JDK 25 did **not** force this; the old `SecurityManagerEx.getClassContext()` works fine on 25. The floor comes from replacing it with `StackWalker` so that no `extends SecurityManager` is left. Staying on 8 is possible if we keep that class — happy to go either way.

### Two pre-existing framework bugs found by the TCK — now split out into #554

Once the TCK could actually launch (see below) it reported 33 errors, all from `FilterImpl.WrapperCapability`. Both were introduced by 466eb93f1c *"[fw] reduce warning related to types Classes"* — a generics cleanup — and are **unrelated to this PR**. They are on master today.

**These are now proposed separately against `master` in #554**, with regression tests, since they affect released versions independently of any Java version work. This branch carries the same fix so that its TCK run can pass; that will drop out once #554 merges.

**1. `Filter.matches(Map)` is broken for any non-empty map.** The constructor lost its assignment into a stray empty `if` block:

```java
m_map = Collections.emptyMap();
if(map != null ) {
}
m_map.putAll(map);
```

`m_map` is an immutable empty map, so `putAll` throws `UnsupportedOperationException` (or NPE when `map` is null). It previously read `m_map = (map == null) ? Collections.EMPTY_MAP : map;`. **This has been broken on master since April 2025.**

**2. `WrapperCapability(ServiceReference)` requires an OSGi 1.10 method.** It was rewritten to `new DictionaryToMap(sr.getProperties(), false)`. `ServiceReference.getProperties()` was only added in Core 1.10 and is not implemented by every `ServiceReference` — the TCK's own mock throws `UnsupportedOperationException`. Restored to the `getPropertyKeys()`/`getProperty()` loop.

See #554 for the details, the fix and the regression tests. A backport may be worth considering, since `Filter.matches(Map)` is unusable in released versions.

### The TCK was failing on every JDK — three stacked causes

`framework.tck` has been red for a while, and none of it was about Java 25:

1. `tck.bndrun` required `assertj-core [3.27.3,3.27.4)` after the dependency was bumped to 3.27.7 in #478, so the bndrun could not resolve at all.
2. With that fixed it resolved but died in the bnd launcher: `aQute.launcher.Launcher.activate()` calls `Policy.setPolicy()` unguarded, which Java 24+ rejects. bnd has since wrapped that in a catch for exactly this reason; the tester was pinned to `biz.aQute.junit` 6.4.1, predating the fix, while the bnd plugins were on 7.1.0. Both moved to **7.4.0**, matching the bndlib version the repository already uses.
3. Then `assertj-core` 3.27.7 could not resolve: it imports `net.bytebuddy [1.18.0,2.0.0)` but byte-buddy was pinned at **1.17.5** — the same omission as (1), from the same dependabot bump. Moved to 1.18.12.

Plus the two `FilterImpl` bugs above. **The OSGi Core R8 TCK now passes on JDK 25** (8.0.0 is the latest published version of `org.osgi.test.cases.framework`, so we are on the current TCK).

Causes 1 and 3 break the TCK on *every* JDK, not just 25, so they are also part of #554 to make that PR verifiable against `master`. Cause 2 stays here, being specific to Java 24 and later.

### CI was reporting failing builds as green

`continue-on-error: true` on the SCR, HTTP and Framework steps made GitHub report those steps as successful even when Maven failed. On one run the "Felix HTTP" step shows `conclusion: success` while its log contains `BUILD FAILURE`. **Removed**, which immediately exposed a real JDK 25 failure it had been hiding:

- `http/jetty` and `http/jetty12` provision **ASM 9.7** bundles into the pax-exam container. ASM 9.7 only understands class files up to Java 23, so `JakartaSpecificWebsocketIT` failed inside Jetty's websocket component with `Unsupported class file major version 69`, taking down the whole http reactor. Now provisions ASM 9.8.

Also fixed: `gogo` was in the `pull_request` trigger but not `push`; `framework.tck` was in neither trigger and had no path filter, so TCK-only changes never triggered a build at all. All modules changed here are now covered, and a `concurrency` group cancels superseded PR runs.

### Other JDK 25 blockers fixed

| | |
|---|---|
| `http/jetty`, `http/jetty12` | provisioned ASM 9.7 into the pax-exam container; 9.7 cannot read class file 69, so websocket ITs failed with `Unsupported class file major version 69`. Now 9.8 |
| `gogo` | mockito-core 5.17.0 to 5.18.0; 5.17's byte-buddy cannot mock on JVM 25 |
| `framework` | `asm-all` 5.2 to `asm`/`asm-tree` 9.8; 5.2 cannot read class files newer than Java 8 |
| `framework.tck` | assertj range, byte-buddy range, bnd 7.4.0, duplicate dependency (see above) |

### Scope

This PR is deliberately limited to the modules that were already in the CI matrix and genuinely broke on JDK 25. Work that turned out to be unrelated has been split into its own pull requests, stacked on this branch:

- **#555** — repairs the builds of `utils`, `connect`, `bundlerepository`, `resolver`, `log.extension` and the `examples/extenderbased` modules. Those cannot be built on a modern JDK, but none of them has a code change, none was in the CI matrix, and each releases independently, so repairing their stale parent poms belongs on its own.
- **#556** — `configadmin`'s `Thread.stop()` to `interrupt()`. `Thread.stop()` has thrown since Java 20, so it is a bug fix in its own right rather than part of Java 25 support.
- **#554** — the `FilterImpl` defects, against `master`, since they affect released versions independently of any Java version work.

`log` was dropped entirely: it needed no change. Its `maven.compiler.source/target` of 1.7 is overridden by `felix-parent`'s `release` configuration, and it builds on JDK 25 unmodified, which was verified.

### Version bumps

Versions follow whether a module's code actually changed:

| module | code changed | version |
|---|---|---|
| `framework` | 16 source files | **7.1.0 to 8.0.0** |
| `scr` | 5 source files | 2.2.19 to 2.3.0 |
| `webconsole` | 3 source files | 5.0.19 to 5.1.0 |
| `gogo.runtime` | 1 source file | 1.1.7 to 1.2.0 |
| `http` | test code only | unchanged |
| `framework.tck`, `main`, `main.distribution` | none | unchanged |

The framework's major bump rests on the removal of the OSGi security layer, a documented behaviour change, and therefore holds **independently of the minimum Java version discussion**. Even keeping the baseline at 8, that removal alone warrants it. Its manifest now reads:

```
Bundle-Version: 8.0.0.SNAPSHOT
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9))"
```

Note that `tck.bndrun` pinned `-runfw` to `[7.0.0,8)`, which would have excluded the new framework, so that range moves to `[8.0.0,9)`.

`scr`, `webconsole` and `gogo.runtime` get **minor** bumps rather than major ones: no exported API differs, and the removed code only ran with a Security Manager installed, which cannot happen on the JDKs this line targets.

`framework.tck`, `main` and `main.distribution` keep their versions, since they have no code changes; only their reference to the framework moves to 8.0.0-SNAPSHOT, which the build requires. Historically these have been released in lockstep with the framework, so if the project prefers to keep that convention they should be bumped too.

No **exported package** versions were changed. Everything the framework exports is `org.osgi.*`, whose versions are set by the specification rather than by us, and the permission API is deliberately kept unchanged, so no exported package's API actually differs.

### Results on JDK 25

`framework` 118 tests, `scr` 170, `webconsole` 23, plus `gogo`, `http`, and the **OSGi Core TCK**, which was re-run and still passes with the framework at 8.0.0. In every module the results **match the JDK 21 control run exactly**. The residual failures are pre-existing platform-specific ones (a test fixture filename containing characters invalid on NTFS, and `deleteDir` file locking) that fail identically on JDK 21 and pass on Linux.

### Deliberately not in scope

- **iPOJO** — 22 poms with `source`/`target` 1.4 to 1.6, left alone since there is a removal PR open.
- **eventadmin** and **bundlerepository.osgi-ct** — not actively maintained. eventadmin is broken independently of JDK: its `maven-bundle-plugin` 4.2.1 bundles a bnd that hits a `TreeMap.computeIfAbsent` `ConcurrentModificationException` (fails identically on **JDK 21**), and its integration tests fetch a bundle from the long-dead `repository.springsource.com`.
- **The wider bnd and maven-bundle-plugin upgrade** — only matters once a module is actually *compiled* at release 25, which none are.
- **Modules outside the CI matrix** — see #555 and #556 above.

### Separate: `sun.misc.Unsafe` and the URL singletons

@tjwatson's point in [#433 (comment)](https://github.com/apache/felix-dev/pull/433#issuecomment-3073468820) is **not** a JDK 25 blocker — the `Unsafe` block is guarded by `catch (Throwable)` with a `setAccessible` fallback, so on 25 it only warns. It is a JDK 26+ and framework-interop concern, so I have kept it out of this PR and opened **#552** as a draft prototype using plurl. There is an unresolved licensing question there (the vendored sources carry EPL-2.0 headers while plurl's LICENSE and pom say Apache-2.0) — @tjwatson, your input would be very welcome on that PR.

paulrutter:
@apache/felix-committers please take a look at https://github.com/apache/felix-dev/pull/433#issuecomment-5464533076 and let me know your thoughts on this. I would like to move this JDK 25 thing forward, given it's released for a while now.

I does bring the baseline for framework to JDK 9, but given we can still maintain a JDK 8 branch if required, i would think that's acceptable.

mbien:
> I does bring the baseline for framework to JDK 9, but given we can still maintain a JDK 8 branch if required, i would think that's acceptable.

JDK 11 as lower bound would be an option too - it will receive updates till next year at least. The libs might be able to use a few new APIs and it might be easier to test, given that running CI against EOL JDK 9 is always a bit awkward.

Projects stuck on EOL JDKs typically don't care about dependency updates, otherwise they wouldn't be there in the first place. So the scenario that someone is running JDK 9 && wants to update felix to the latest version is quite unlikely IMO. (I am no committer)

paulrutter:
> > I does bring the baseline for framework to JDK 9, but given we can still maintain a JDK 8 branch if required, i would think that's acceptable.
> 
> JDK 11 as lower bound would be an option too - it will receive updates till next year at least. The libs might be able to use a few new APIs and it might be easier to test, given that running CI against EOL JDK 9 is always a bit awkward.
> 
> Projects stuck on EOL JDKs typically don't care about dependency updates, otherwise they wouldn't be there in the first place. So the scenario that someone is running JDK 9 && wants to update felix to the latest version is quite unlikely IMO. (I am no committer)

That's a fair point, but given what i know from other PRs where the baseline was proposed to be lifted, we got some pushback on it. I agree that consumers on older JDK versions can remain to use the 7.x line, while 8.x should lift the baseline.

mbien:
> That's a fair point, but given what i know from other PRs where the baseline was proposed to be lifted, we got some pushback on it.

I think it will depend on how the lower bound changes though. If its in the next major version while the last major version is still maintained - i can't see how this would cause pushback.

My main point was that bumping the lower bound from LTS 8 to EOL 9 (on the next major version) is not much different to bumping it from LTS 8 to LTS 11 - since it isn't 8 anymore in both cases ;)

PR #458, opened 2025-11-24
FELIX-6812: Fix ConfigInstaller to use listConfigurations instead of getConfiguration

FelixMarxIBM:
Replace `getConfiguration()` call with `listConfigurations()` in `ConfigInstaller.configurationEvent()` method to avoid unintentionally creating new configurations. The `getConfiguration()` method creates a configuration if it doesn't exist, which is not the desired behavior when simply checking for existing configurations.

Changes:
- Use `listConfigurations("(service.pid=" + pid + ")")` instead of `getConfiguration(pid, "?")`
- Add proper null and length checks for configuration arrays
- Update corresponding test mocks to reflect the corrected API usage

This fixes an issue with felix fileinstall reacting upon the CM_UPDATED events and creating a configuration that was already deleted.


stbischof (approved):
Lgtm

cziegeler (approved):

PR #467, opened 2026-01-21
FELIX-6815 Fix race condition for defaultBaseUrl

ghenzler:


laeubi (commented):
Instead of a raw service listener I would use a `ServiceTracker` instead what has a method `getTracked()` what gives you both the service and the servicereference to check the properties from or you use a customizer that holds the interesting properties in an AtomicReference.

PR #506, opened 2026-05-27
FELIX-6844: Implement self-healing recovery mechanism for FelixDispatchQueue thread in EventDispatcher

sahvx655-wq:
This PR introduces a controlled self-healing mechanism for the FelixDispatchQueue thread in EventDispatcher to prevent permanent asynchronous event failure when the dispatcher thread crashes due to severe runtime errors (e.g., OutOfMemoryError, StackOverflowError).

Currently, if the dispatcher thread dies, the framework permanently stops processing asynchronous events. This change adds a safe recovery strategy while preserving the existing single-threaded design.

### DIFF
--- a/framework.tck/pom.xml
+++ b/framework.tck/pom.xml
@@ -105,7 +105,7 @@
     <dependency>
       <groupId>net.bytebuddy</groupId>
       <artifactId>byte-buddy</artifactId>
-      <version>1.17.5</version>
+      <version>1.18.0</version>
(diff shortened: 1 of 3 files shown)


sahvx655-wq:
CI is failing due to AssertJ version range mismatch: required [3.27.3,3.27.4) but 3.27.7 is available in the resolver. Should the range be widened in tck.bndrun?

PR #508, opened 2026-05-28
FELIX-6831 Use Jetty EE10 artifacts instead of EE11 on Jetty 12.1.9

paulrutter:
https://issues.apache.org/jira/browse/FELIX-6831

- pom.xml: org.eclipse.jetty.ee11(.websocket) groupIds -> ee10, jetty-ee11-* artifactIds -> jetty-ee10-*, bnd Export-Package and groovy-magic skip list updated to ee10.
- JettyService / JettyErrorHandler: switch imports and log messages from ee11 to ee10 servlet/websocket packages.
- PaxExam integration tests: ee11 maven bundle coords -> ee10.


paulrutter:
@cziegeler what do you think about maintaining a branch like this?
It would still use the latest and greatest Jetty version, but then the EE10 bundles are used instead.

This branch will not be merged to master, but be kept in sync with it (sans the EE11 imports).

cziegeler:
yes, makes sense to me

paulrutter:
> yes, makes sense to me

Thanks, then i will move forward with releasing this as well in the near future.
Closing PR as this won't be merged to master.

PR #509, opened 2026-05-28
FELIX-6832 Fix Jetty 12.1 deprecation warnings

paulrutter:
## Summary

Fixes the seven deprecation warnings emitted by the `http/jetty12` module against Jetty 12.1.x:

- `VirtualThreadPool.setMaxThreads(int)` / `getMaxThreads()` → `setMaxConcurrentTasks(int)` / `getMaxConcurrentTasks()` ([JettyService.java](src/main/java/org/apache/felix/http/jetty/internal/JettyService.java)).
- `org.eclipse.jetty.server.handler.gzip.GzipHandler` → `org.eclipse.jetty.compression.server.CompressionHandler` + `GzipCompression` + `CompressionConfig` (from the new `jetty-compression-server` / `jetty-compression-gzip` modules). All include/exclude lists (methods, paths, mime types), min compress size, and sync flush are preserved through the new builder / encoder-config APIs.
- Replaced the previously-public `GzipHandler.DEFAULT_MIN_GZIP_SIZE` constant (the new `GzipCompression` keeps that field private) with a local `JettyConfig.DEFAULT_GZIP_MIN_SIZE = 32`.

Jira: [FELIX-6832](https://issues.apache.org/jira/browse/FELIX-6832)

## Notable behavior change

The `org.apache.felix.jetty.gzip.inflateBufferSize` property is **removed**. The new `CompressionHandler` API has no equivalent knob, so the property no longer had any effect. Leaving a dead config option was misleading, so the constant, the `JettyConfig.getGzipInflateBufferSize()` getter, and the metatype entry were all dropped. Existing OSGi configs that set this property will be silently ignored.

Decompression of incoming requests is now governed by `CompressionHandler`'s defaults rather than the old `inflateBufferSize > 0` opt-in.

## Test plan

- [x] `mvn clean compile` on `http/jetty12` — no deprecation warnings remain (compared with the seven warnings on master).
- [x] `mvn test-compile` on `http/jetty12` — succeeds.
- [x] Run pax-exam ITs.
- [ ] Smoke-test a deployment with `org.apache.felix.jetty.gziphandler.enable=true` to confirm gzip compression still works end-to-end.


paulrutter:
@cziegeler i tried to resolve the deprecation warnings, which meant dropping `inflateBufferSize` as there is no longer a API counterpart.

cziegeler (approved):
Thanks, looks good and makes sense to remove the unused config option completely

PR #510, opened 2026-05-30
FELIX-6841: Fix ArrayIndexOutOfBoundsException in Pipe.isTty

sahvx655-wq:
Fixes an `ArrayIndexOutOfBoundsException` in `Pipe.isTty(int fd)` when the file descriptor index `fd` is equal to or greater than the `streams` array length.

The boundary check in `Pipe.isTty` used `fd > streams.length` instead of `fd >= streams.length`. Because Java arrays are 0-indexed, checking `fd` at `streams.length` bypassed the check and threw an `ArrayIndexOutOfBoundsException` when accessing `streams[fd]`.

- Updated the check to `fd >= streams.length` to safely return `false` on out-of-bounds descriptors.
- Added `testIsTtyOutOfBounds` in `TestParser.java` and verified the build passes successfully.


sahvx655-wq:
any update?

PR #511, opened 2026-05-30
FELIX-6842: gogo-jline: Implement support for -u (UTC) option in date command

sahvx655-wq:
This PR fixes handling of the documented `-u` (UTC) timezone option in the JLine shell `date` command.

In `org.apache.felix.gogo.jline.Posix.java`, the `date` command usage documentation includes the `-u` option (use UTC), but the argument parsing logic did not recognize it and instead fell through to the default error path, resulting in an `IllegalArgumentException`.

 Detect the `-u` option during argument processing.
 Treat `-u` as a valid option.
 Apply the UTC timezone when parsing or formatting dates.

Added `testDateUtc` in `PosixTest.java`.
Verified the new test passes successfully.


sahvx655-wq:
any update?

PR #512, opened 2026-06-01
FELIX-6840: Harden XML parsers against XML External Entity (XXE) injection

sahvx655-wq:
This PR hardens XML parsing across Apache Felix by disabling DTD processing, external entity resolution, and XInclude support in SAX, DOM, StAX, and XMLReader implementations used by core components, libraries, and build tools.

Parser Hardening:Secured XML parser factories in Declarative Services (SCR), iPOJO Manipulator, OSGi Check Maven Plugin, and StAX parsers.
XXE Validation: Added `testParserRejectsXXE` in `XmlHandlerTest` to verify that XML containing external entities or DOCTYPE declarations is safely rejected.
Compatibility: Existing OSGi descriptors do not rely on DTDs or external entities, so the changes are expected to be backward-compatible.


sahvx655-wq:
any update?

PR #513, opened 2026-06-01
FELIX-6839: Resolve lifecycle race condition in EventDispatcher start/stop sequence

sahvx655-wq:
This PR addresses a race condition in `EventDispatcher` that can occur during concurrent framework start/stop operations in the same JVM.

 Updated `startDispatching()` to wait while the dispatcher is being stopped.
 Ensures shutdown completes before a new dispatch thread is created, preventing inconsistent dispatcher lifecycle states.


sahvx655-wq:
any update?

PR #514, opened 2026-06-02
FELIX-6838: Fix Platform-Dependent URI Mock Failures in fileinstall Tests

sahvx655-wq:
Fixes failures in `ConfigInstallerTest` and `DirectoryWatcherTest` caused by platform-dependent URI construction in mock expectations.

The tests previously constructed expected file URIs by manually concatenating strings (for example, prefixing paths with `"file:"`). This approach can produce URI formats that differ across operating systems and may not match the URIs generated by the implementation.

Replace manual URI construction with `File.toURI()` to ensure test expectations use the same platform-independent URI representation as the production code, eliminating OS-specific test failures.


sahvx655-wq:
gentle ping

PR #515, opened 2026-06-02
FELIX-6837: Skip Write-Protection Test on Unsupported File Systems

sahvx655-wq:
Modifies `testInvalidTempDir` in `DirectoryWatcherTest` to verify that `parent.setWritable(false)` actually takes effect before proceeding with the test.

Some file systems and operating systems do not reliably enforce directory write-permission changes (for example, common Windows/NTFS configurations for the directory owner). In such environments, the test's assumptions are invalid and can lead to false failures.

If write access cannot be revoked as expected, the test exits early instead of failing, ensuring consistent behavior across platforms while preserving test coverage where permission enforcement is supported.


sahvx655-wq:
any update?

PR #516, opened 2026-06-02
FELIX-6836: Mitigate Zip Slip / Path Traversal (CWE-22) Vulnerability in Archive Extraction and Cache Handlers

sahvx655-wq:
This PR prevents Zip Slip / Path Traversal issues during ZIP and JAR archive extraction by validating archive entry paths before writing files to disk. Entries that would resolve outside the intended extraction directory are rejected, ensuring extraction remains confined to the target location.

In addition to the implementation changes, regression tests have been added to verify that malicious archives containing traversal entries (for example, `../../evil.txt`) are rejected and cause extraction to fail safely.

### DIFF
--- a/bundlerepository/pom.xml
+++ b/bundlerepository/pom.xml
@@ -107,8 +107,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <target>1.5</target>
-          <source>1.5</source>
+          <target>1.8</target>
+          <source>1.8</source>
         </configuration>
       </plugin>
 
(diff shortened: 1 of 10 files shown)


paulrutter:
@sahvx655-wq before you start creating more PR's, let's first establish what the problem is you're trying to solve.
Every PR should be accompanied with a Jira ticket in https://issues.apache.org/jira/projects/FELIX/summary.

Can you start there first and create tickets for the things you're trying to address?
I want to prevent having PR's that no-one exactly knows the reason for.

sahvx655-wq:
I have created the Jira ticket for this issue here : https://issues.apache.org/jira/browse/FELIX-6836

sahvx655-wq:
any update?

PR #518, opened 2026-06-04
FELIX-6835 - CommentRemovingReader incorrectly handles escaped quotes and drops characters at buffer boundaries

rombert:


cziegeler (approved):

rombert:
Thanks for the review @cziegeler . Can you please also add the `cm.json-2.0.8` fix version to https://issues.apache.org/jira/browse/FELIX-6835 ? It is missing.

PR #543, opened 2026-08-11
Restore Apache Felix Logback module

wilx:
I would like to restore the Logback component. I have restored the original source and updated it with Logback 1.5.x, plus some changes I see as beneficial. It works for me in production like this.


rombert:
Speaking as a user of Felix Logback that would IMO be good. I am a Felix committer but never worked with the logback bundle.  What do other Felix committers/PMC members think ( /cc @apache/felix-committers )  ?

cziegeler:
We removed this couple of years ago due to lack of maintainers . We had a similar situation with another Apache Felix module where we actually reverted the removal, added a new committer promising to help on that module. And it resulted in a larger unmaintained code base, as after an initial fix nothing else happened.
I dont want to repeat that - the code is under the ASL 2 so it is easy to "continue" it somewhere else.
Now, if we decide to bring it back in, I suggest we directly revert the removal commit - and not do it via a PR

wilx:
What maintenance does this need beside keeping it on par with some Logback releases?

cziegeler:
Right, keeping all dependencies up to date, releasing whenever required - fixing issues if they get reported.
It is not a long list, but nevertheless it needs someone to do it regularly and not just for one release

PR #548, opened 2026-08-24
cm.json: Improve Configurator specification conformance

wilx:
This updates `cm.json` to Felix parent 9 and tightens its handling of OSGi Configurator JSON resources.

The changes:

- convert mixed JSON arrays to `String[]` as required by Configurator [§150.3.4](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.configurator.html#d0e130545);
- strictly validate `:configurator:resource-version`, `:configurator:symbolic-name`, and `:configurator:version` according to [§150.3.1](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.configurator.html#d0e130346), including metadata supplied in bundle resources;
- preserve PID-level Configurator property keys so suffixed aliases are not interpreted as the standard policy or ranking keys, as defined in [§150.3.3](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.configurator.html#d0e130482); and
- add regression coverage before each behavioral fix.

The downstream Configurator module is also updated to Felix parent 9 and from `org.apache.felix.cm.json` 2.0.0 to this branch's 2.0.9-SNAPSHOT. This makes its Java 17 build use the current `cm.json` implementation.

The metadata work intentionally does not change how PID-level `:configurator:` properties are retained when no property handler is installed; that can be addressed separately with an API design suitable for round-tripping.

Validation on Java 17:

```
env JAVA_HOME=/opt/jdks/latest-17 nice -n 10 /opt/maven/latest/bin/mvn -B -V -Dstyle.color=always -f cm.json/pom.xml clean install
env JAVA_HOME=/opt/jdks/latest-17 nice -n 10 /opt/maven/latest/bin/mvn -B -V -Dstyle.color=always -f configurator/pom.xml clean verify
```

Results:

- `cm.json`: 60 tests, 0 failures/errors; bundle baseline, RAT, and legal-file checks pass.
- Configurator: 18 tests, 0 failures/errors; bundle creation, RAT, and legal-file checks pass.

### DIFF
--- a/cm.json/pom.xml
+++ b/cm.json/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>org.apache.felix</groupId>
         <artifactId>felix-parent</artifactId>
-        <version>7</version>
+        <version>9</version>
         <relativePath />
     </parent>
 
--- a/cm.json/src/main/java/org/apache/felix/cm/json/io/impl/ConfigurationReaderImpl.java
+++ b/cm.json/src/main/java/org/apache/felix/cm/json/io/impl/ConfigurationReaderImpl.java
@@ -20,12 +20,14 @@ package org.apache.felix.cm.json.io.impl;
 
 import java.io.IOException;
 import java.io.Reader;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 
 import jakarta.json.JsonException;
+import jakarta.json.JsonNumber;
 import jakarta.json.JsonObject;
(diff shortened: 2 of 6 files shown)


stbischof:
would you be so nice and test to the osgi tck?
in max 1 week you kan find it here
https://github.com/osgi/org.osgi.service.configurator

wilx:
I have created https://github.com/wilx/felix-configurator-tck project to test against the test in OSGi TCK that exercise it. The branch passes. If this is what you meant.

wilx:
Ping. Was the linked repo supposed to contain something?

